PR #4, opened 2018-11-02
Continuous integration setup

cnovak:
Continuous integration using [deviantintegral/drupal_tests repo](https://github.com/deviantintegral/drupal_tests)

Custom work was done in Robo file configureM10nDependencies() method to pull in all needed dependencies, pulling in apigee-client-php via source and from chapter3 repo mint branch and pulling in apigee-edge-drupal module via source.  Also needed to add commerce and token modules to get tests to run

Example run: https://circleci.com/workflow-run/9cbf6200-eb43-4c18-9d94-f6f332177bb1

- Codesniffer is failing due to issues that need to be fixed in code, will be fixed w/separate PR
- Functional tests are failing due to `Exception: Error: Maximum function nesting level of '256' reached, aborting!` error to be fixed w/separate PR


cnovak:
Got error:
```
Drupal\Tests\apigee_edge\Functional\DeveloperAppFieldTest::testFieldableDeveloperApp
Behat\Mink\Exception\ExpectationException: The string "These settings apply to the <em class="placeholder">RVG4CIWZ</em> field everywhere it is used." was not found anywhere in the HTML response of the current page.
/var/www/html/vendor/behat/mink/src/WebAssert.php:770
/var/www/html/vendor/behat/mink/src/WebAssert.php:325
/var/www/html/build/core/tests/Drupal/FunctionalTests/AssertLegacyTrait.php:312
/opt/drupal-module/.travis/build/core/modules/field_ui/src/Tests/FieldUiTestTrait.php:45
/opt/drupal-module/tests/src/Functional/DeveloperAppFieldTest.php:265
/opt/drupal-module/tests/src/Functional/DeveloperAppFieldTest.php:120
```
will try restarting once more.

cnovak:
I accidentally restarted the whole build, so a few jobs are failing again

.circleci/config.yml
```diff
@@ -0,0 +1,244 @@
+# Default configuration file for Drupal modules.
```
Jaesin, line 1: We might want to replace this readme info with something more useful such as an overview of the setup and links to libraries/tools being used..

.circleci/config.yml
```diff
@@ -0,0 +1,244 @@
...
+#      any private repositories.
+#   6. Create a status badge embed code in Circle and add it to the README.md.
+#
+# Check https://circleci.com/docs/2.0/language-php/ for more details
+#
+
+defaults: &defaults
+  docker:
+    # specify the version you desire here (avoid latest except for testing)
+    - image: andrewberry/drupal_tests:0.4.0
```
Jaesin, line 19: This is using `drupal:8.6-apache` which is php 7.2 do we want to have the ability to test against multiple versions of PHP?

.circleci/config.yml
```diff
@@ -0,0 +1,244 @@
...
+        at: /var/www/html
+
+    - checkout
+
+    - run:
+        working_directory: /var/www/html
+        command: |
+          cp ./modules/apigee_m10n/.circleci/update-dependencies.sh /var/www/html
+          cp ./modules/apigee_m10n/.circleci/RoboFile.php /var/www/html
+          ./code-sniffer.sh apigee_m10n
```
Jaesin: I think this is using the standard drupal rule set. Should we make it use the `phpcs.xml.dist` that is included with the module?

mxr576:
@cnovak Why don't you use the same Travis CI based setup for this module as the Apigee Edge module uses? Why do you want to re-invent the wheel here?

.circleci/config.yml
```diff
@@ -0,0 +1,244 @@
...
+          cp ./modules/apigee_m10n/.circleci/update-dependencies.sh /var/www/html
+          cp ./modules/apigee_m10n/.circleci/RoboFile.php /var/www/html
+          ./test-functional-js.sh apigee_m10n
+
+    - store_test_results:
+        path: /var/www/html/artifacts/phpunit
+    - store_artifacts:
+        path: /var/www/html/artifacts
+
+# Run Behat tests. This command invokes the behat.sh hook.
```
Jaesin: We don't have any Behat tests or plans to implement them.

.circleci/config.yml
```diff
@@ -0,0 +1,244 @@
...
+          ./test.sh apigee_m10n
+
+    - store_test_results:
+        path: /var/www/html/artifacts/phpunit
+    - store_artifacts:
+        path: /var/www/html/artifacts
+
+# Run Drupal functional tests. This command invokes the test-functional.sh
+# hook.
+functional_tests: &functional_tests
```
Jaesin, line 105: I'm wondering if WebDriver or the GoutteDriver is being used form functional tests.

tests/src/Behat/example.behat.local.yml
```diff
@@ -0,0 +1,31 @@
+# Local Behat configuration file.
```
Jaesin: This is only useful if we really work on the Behat test suite. I would leave this out unless the testing plan changes.

.circleci/config.yml
Jaesin: I found it. It would default to GoutteDriver as it isn't overridden in <https://github.com/deviantintegral/drupal_tests/blob/master/hooks/test-functional.sh>.

.circleci/config.yml
```diff
@@ -0,0 +1,244 @@
...
+          ./test-functional.sh apigee_m10n
+
+    - store_test_results:
+        path: /var/www/html/artifacts/phpunit
+    - store_artifacts:
+        path: /var/www/html/artifacts
+
+# Run Drupal functional tests. This command invokes the test-functional-js.sh
+# hook.
+functional_js_tests: &functional_js_tests
```
Jaesin, line 127: Note: This is going to test with chrome using selenium server. See: https://github.com/deviantintegral/drupal_tests/blob/master/hooks/test-functional-js.sh

Jaesin:
There are currently some tests that fail because we haven't been using `GoutteDriver` for functional tests because it's  much slower than ChromeDriver. There is a task to fix them so they pass with `GouteDriver`.

cnovak:
> @cnovak Why don't you use the same Travis CI based setup for this module as the Apigee Edge module uses? Why do you want to re-invent the wheel here?

@mxr576 I found an [article by Lullabot](https://www.lullabot.com/articles/continuous-integration-drupal-8-circleci) that said "if you are building a Drupal module, check out [drupal_tests](https://github.com/deviantintegral/drupal_tests)" and wanted to check it out.  

I did not have to spend much time on it at all since the setup was very quick, and as you can see I did not do many changes to it.  I wanted to put it up and see how it compares to the travisCI setup we already have.  I did not have to reinvent the wheel, just used someone else's wheel. 

What I like about it:
1. The [test failures are very clear](https://circleci.com/gh/cnovak/apigee-m10n-drupal/77), no hunting to figure out what went wrong.
2. [Test coverage included](https://78-154601551-gh.circle-artifacts.com/0/var/www/html/artifacts/coverage-html/index.html)
3. We are using a testing framework that other Drupal modules use, allowing us to collaborate with others on fixing issues and maintaining the testing setup
4. RoboFiles make the testing easy to understand

What is missing:
1. Does not test multiple versions of dependencies and PHP versions
2. I added configureM10nDependencies() to the RoboFile.  Maybe that should be in a separate file to make upgrades easier.  Might not need that command after everything is up on Drupal.org however.
3. Did not add enough info/docs to explain what I did as @Jaesin pointed out

Jaesin:
It looks like the current tests are failing because of `xdebug.max_nesting_level = 256` 8192 seems to be enough. You can set it for just the process with `php -d xdebug.max_nesting_level=8192 vendor/phpunit/phpunit/phpunit...`

Jaesin:
That said, enabling xdebug during testing slows things down. I think it's only required for the coverage report.

Jaesin (approved):
Besides updating the comment in the CI file, this should  be ready to go.

rspier:
@googlebot Ping

cnovak:
@Jaesin updated documentation, let me know if you can think of any info I missed.

PR #8, opened 2018-11-30
[DRUP-370] auth changes

Jaesin:
Since `apigee_edge` 8.x-1.x-beta1 was released, the key module plugins we are using have been deprecated and removed.

Commit:563b134 removes an unnecessary try/catch that was added in commit:8fa5094 and queues the correct response.


minnur:
Looks good.

minnur (approved):
Looks good.

PR #12, opened 2018-12-08
Fix route that is using a string instead of account interface.

Jaesin:


src/Controller/BillingController.php
```diff
@@ -94,23 +93,15 @@ public function myPrepaidBalance(): RedirectResponse {
...
    * @throws \Exception
    */
-  public function prepaidBalancePage(string $user) {
-    $user = User::load($user);
-
-    if (!$user) {
-      throw new NotFoundHttpException();
-    }
-
+  public function prepaidBalancePage(AccountInterface $user) {
```
dawehner: I would have expected that you have to typehint to `UserInterface` to trigger the automatic parameter conversion.

Jaesin: Good catch. I changed it to use `UserInterface`.

dawehner (approved):
Looks great!

minnur:
Looks good

minnur (approved):
Looks good

PR #23, opened 2019-01-03
[DRUP-508] Access is required to view rate plans.

Jaesin:


tests/src/Kernel/Controller/PackageControllerKernelTest.php
```diff
@@ -54,12 +55,17 @@ protected function setUp() {
     $this->installConfig([
       'user',
     ]);
+    // Install user 0 and user 1.
+    \user_install();
```
minnur, line 59: Maybe remove `slash` at the beginning of the function name ? Is it really needed ?

tests/src/Traits/ApigeeMonetizationTestTrait.php
```diff
@@ -144,7 +144,7 @@ protected function createAccount(array $permissions = [], bool $status = TRUE, s
       'first_name' => $this->randomMachineName(),
       'last_name' => $this->randomMachineName(),
       'name' => $this->randomMachineName(),
-      'pass' => user_password(),
+      'pass' => \user_password(),
```
minnur, line 147: Is there a reason for a slash at the beginning of the function ?

minnur (commented):
Few small questions/clarifications.

tests/src/Traits/ApigeeMonetizationTestTrait.php
Jaesin: When the user module isn't enabled yet, php is looking for `Drupal\Tests\apigee_m10n\Traits\user_password()`.

minnur (approved):
Looks good.

PR #24, opened 2019-01-04
[DRUP-523], [DRUP-507] Not accepting T&C on plan puts error on top of page, disru…

minnur:


cnovak:
@minnur looks like this PR should be closed because #25 has same changes?

minnur:
@cnovak you're right :( I should've created branch off 8.x-1.x, instead I created off current working. Since the changes are minimal that shouldn't be a problem.

Jaesin:
While testing manually, I clicked purchase without checking accept, TnC. I was redirected to the form but no error showed. Then, when I click again without checking TnC, This is what I get: 
image

minnur:
Lets close this PR in favor of https://github.com/apigee/apigee-m10n-drupal/pull/24

Jaesin:
Did we try making TnC required in `\Drupal\apigee_m10n\Entity\Subscription::baseFieldDefinitions()` via `$definitions['termsAndConditions']->setRequired(TRUE);`? This would set `required="required"` on the checkbox HTML element preventing the form from being submitted unless checked.

Jaesin:
I am wondering if something needs to be done to auto accept my no-reply email address for the CLA.

cnovak:
I looked up both 'Jaesin' and 'minnur' accounts, both have accepted CLA

src/Entity/Form/SubscriptionForm.php
```diff
@@ -67,6 +68,16 @@ public function getFormId() {
     return parent::getFormId() . '_' . $this->entity->getRatePlan()->id();
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function form(array $form, FormStateInterface $form_state) {
+    $form = parent::form($form, $form_state);
+    // Redirect to Rate Plan detail page on submit.
+    $form['#action'] = $this->getEntity()->getRatePlan()->url();
```
Jaesin: I think this should be `$form['#action'] = $this->getEntity()->getRatePlan()->url('subscribe');` because the canonical view display could be configured to show a link to the subscribe form instead of the rendered form.

Jaesin (approved):
I think this can go forward. There is one followup TODO.

PR #26, opened 2019-01-09
[DRUP-401] packge entities

Jaesin:
Creates a package entity to use instead of accessing the packages controller directly.


minnur (approved):
The code looks good to me,  I haven't had a chance to run it.

tests/src/Kernel/TestFrameworkKernelTest.php
```diff
@@ -109,10 +109,10 @@ public function testIntegrationToggle() {
       ->getConfig('handler');
 
     if ($this->integration_enabled) {
-      self::assertTrue(($handler instanceof HandlerStack));
+      self::assertInstanceOf(MockHandler::class, $handler);
```
shadcn, line 112: This looks good. All tests green now.

Jaesin: This wrong and failing for me. If integration is enabled, the mock handler should  be disabled. That means the handler should be `HandlerStack::class` so we have these backwards.

PR #28, opened 2019-01-11
[DRU-401] Add a patch to fix a core issue where `RevisionableInterfac…

Jaesin:
This patch is needed in order to render any entity that doesn't implement `RevisionableInterface`.

For monetization entities re add `isDefaultRevision` and just return `TRUE` but edge entities don't and that breaks how entity references are rendered.


minnur (approved):
Looks good to me.

PR #29, opened 2019-01-11
[DRUP 385] allows skip cart for product types

shadcn:


modules/add_credit/src/AddCreditService.php
```diff
@@ -161,10 +162,30 @@ public function formCommerceProductTypeEditFormAlter(&$form, FormStateInterface
       '#title' => t('Enable <em>Apigee Monetization Add Credit</em> for this product type.'),
       '#default_value' => $default_value,
     ];
+
+    /** @var \Drupal\commerce_product\Entity\ProductTypeInterface $product_type */
+    $default_value = (($product_type = $form_state->getFormObject()->getEntity())
```
Jaesin: `$product_type` is already defined on line 155.

Maybe we should change it to something like:

```
    $product_type = $form_state->getFormObject()->getEntity();

    // Add an option to allow enabling Apigee add credit for a product type.
    $form['apigee_m10n_enable_add_credit'] = [
      '#type' => 'checkbox',
      '#title' => t('Enable <em>Apigee Monetization Add Credit</em> for this product type.'),
      '#default_value' => $product_type->getThirdPartySetting('apigee_m10n_add_credit', 'apigee_m10n_enable_add_credit'),
    ];

```

etc...

modules/add_credit/src/Form/ApigeeAddCreditAddToCartForm.php
```diff
@@ -0,0 +1,87 @@
...
+   * @return bool
+   *   TRUE is should skip cart.
+   */
+  protected function shouldSkipCart() {
+    /** @var \Drupal\commerce_order\Entity\OrderItemInterface $order_item */
+    $order_item = $this->entity;
+    /** @var \Drupal\commerce\PurchasableEntityInterface $purchased_entity */
+    $purchased_entity = $order_item->getPurchasedEntity();
+
+    try {
```
Jaesin: I don't think you need a try catch here.
```
  protected function shouldSkipCart() {
    return isset($this->entity->purchased_entity->entity->type->entity)
      ? $this->entity->purchased_entity->entity->type->entity->getThirdPartySetting('apigee_m10n_add_credit', 'apigee_m10n_enable_skip_cart', FALSE)
      : FALSE;
  }
```

Jaesin: Maybe it's not that easy, this get's the product variation type. How do we know the product variation bundle ID will be the same as the Product bundle ID? Is it always the same?

Jaesin: You should be able to use: `$this->entity->purchased_entity->entity->product_id->entity->type->entity`

Jaesin (approved):
This looks good to go.

PR #32, opened 2019-01-15
[DRUP-429] Removed route for company profile.

Jaesin:


shadcn (approved):
Looks good 👍

Jaesin:
This actually looks like it has some regressions.

PR #37, opened 2019-01-19
[DRUP-541] Package view mode

Jaesin:
This makes the package view mode configurable and removes the setting from rate plans.


minnur (approved):
Looks good

PR #38, opened 2019-01-23
[DRUP-526] Update tests to use the latest dev version of `apigee_m10n`

Jaesin:
Some changes to the Apigee Edge module are causing the tests to fail. These fixes will ensure the tests pass using the latest or beta1


tests/src/Kernel/Controller/PackageControllerKernelTest.php
```diff
@@ -170,7 +189,10 @@ public function testPackageController($uid, $user_has_access) {
 
     foreach ($sdk_packges as $package) {
```
shadcn: nit: s/sdk_pakges/sdk_packages

shadcn (approved):
This looks good. Confirmed all tests passing with apigee_edge `8.x-1.x`. LGTM.

PR #39, opened 2019-01-28
[DRUP-545] Plan form should display "already purchased" if plan is currently or future active.

minnur:


shadcn (approved):
I fixed the tests and can confirm the functionality for purchased plans is in place. Approved.

PR #43, opened 2019-02-05
[WIP] [DRUP 542] minimum price validation

shadcn:
This refactors the apigee_price_range base field into a configurable bundle field type. The field type ships with two widgets: the default price widget and the price range widget.

top_up_amount_field

All (most) validations / constraints are hence tied to this field instead of altering the commerce price field. This solves the issue with configurable currencies as well.

No more `hook_field_widget_form_alter` to handle the default price on the unit price field either.

_This is WIP. Go test bots!_


modules/add_credit/src/Element/TopUpAmountRange.php
```diff
@@ -197,7 +193,7 @@ public static function getNumberFields() {
     return [
       'minimum' => t('Minimum'),
       'maximum' => t('Maximum'),
-      'default' => t('Default'),
+      'number' => t('Default'),
```
Jaesin, line 196: Would `amount` or `price` be better than number?

modules/add_credit/src/Plugin/Field/FieldType/TopUpAmountItem.php
```diff
@@ -0,0 +1,171 @@
...
+
+/**
+ * Plugin implementation of the 'apigee_top_up_amount' field type.
+ *
+ * @FieldType(
+ *   id = "apigee_top_up_amount",
+ *   label = @Translation("Apigee top up amount"),
+ *   description = @Translation("Stores a top up amount with minimum, maximum and default values."),
+ *   default_widget = "apigee_top_up_amount_price",
+ *   default_formatter = "commerce_price_default",
```
Jaesin, line 37: No available formatter appears for view display

Jaesin (commented):
Pruduct setup from scratch:

#### Add a new product type.

- Label: "Add credit"
- [X] Enable Apigee Monetization Add Credit for this product type.

Edit URL: <http://example.com/admin/commerce/config/product-types/add_credit/edit>

#### Edit the automatically created product variaiton type:
 
<http://example.com/admin/commerce/config/product-variation-types/add_credit/edit/fields>

- [Add](http://example.com/admin/commerce/config/order-item-types/default/edit/fields/add-field) an Apigee top up about field
- [Change](http://example.com/admin/commerce/config/order-item-types/default/edit/form-display) the field widget to "Top up amount (range)".
- Remove the price from the form display.
- Remove the list price from the form display.

#### Configure the order type

<http://example.com/admin/commerce/config/order-item-types/default/edit/form-display/add_to_cart>

-  Set Purchased entity to "Product variation title"
-  Enable the "Unit price" field.

#### Add a product

<http://example.com/product/add/add_credit>

- Try to add a minumum amount that is too low.

        The minimum top up amount for USD is USD11.00.
    
- Setting the Add to cart unit price to "Price" still allows user input for price but doesn't use the input value.

modules/add_credit/src/Element/TopUpAmountRange.php
shadcn: `number` is from the default price field type. So using `number` here makes reusing some code / widgets easy.

cnovak:
It would be good to keep the functionality the same as before this PR since the setup/config is easier for the end user.  I am also concerned that:
1. If the configuration is not followed correctly we will still have a situation where the validation is missed.  
2. The validation is still being done during configuration which could be out of sync if an admin makes modifications on the Apigee Edge UI and forgets to update the configuration in Drupal.  

Instead of accepting this PR, let's add a validation during checkout just to make sure the charge will be higher than the minimum top up amount set in Edge.

shadcn:
Do we want the validation on checkout or when the product is added to cart?

cnovak:
@Jaesin and I discussed "checkout" since we just want to verify the minimum topup was not changed on Edge before we process a credit card.  If someone modifies minimum topup on Edge, we want to detect that right before we start collecting credit card info to make sure the balance update on Edge does not fail.

shadcn:
@cnovak ok. Let's close this PR. I'll open a new one with the implementation on checkout.

PR #51, opened 2019-02-14
[DRUP-542] add minimum top up amount validation on checkout

shadcn:
Adds minimum top amount validation on checkout. WIP.

### DIFF
--- a/modules/add_credit/apigee_m10n_add_credit.module
+++ b/modules/add_credit/apigee_m10n_add_credit.module
@@ -55,58 +55,7 @@ function apigee_m10n_add_credit_entity_bundle_field_info(EntityTypeInterface $en
  * Implements hook_form_alter().
  */
 function apigee_m10n_add_credit_form_alter(&$form, FormStateInterface $form_state, $form_id) {
-  if (($flow = $form_state->getFormObject())
-    && ($flow instanceof CheckoutFlowBase)
-    && ($form['#step_id'] == 'review')
-  ) {
-    // Add a custom validation handler to check for add credit products.
-    array_unshift($form['#validate'], 'apigee_m10n_add_credit_checkout_form_review_validate');
-  }
-}
-
-/**
- * Validation callback for the checkout pane form.
- */
-function apigee_m10n_add_credit_checkout_form_review_validate(array $form, FormStateInterface $form_state) {
-  if ($flow = $form_state->getFormObject()) {
-    // Loop through all order items.
-    // Find the add_credit products and the total amount per currency.
-    /** @var \Drupal\commerce_order\Entity\OrderItemInterface $item */
-    $add_credit_totals = [];
-    foreach ($flow->getOrder()->getItems() as $item) {
-      /** @var \Drupal\commerce_product\Entity\ProductInterface $product */
-      if (($product = $item->getPurchasedEntity()->getProduct())
-        && ($product->hasField('apigee_add_credit_enabled'))
-        && ($product->get('apigee_add_credit_enabled')->value)) {
-        // Use the total price since this factors in the quantity.
-        $price = $item->getTotalPrice();
-        $currency_code = strtolower($price->getCurrencyCode());
-        $add_credit_totals[$currency_code] = !empty($add_credit_totals[$currency_code]) ? $add_credit_totals[$currency_code]->add($price) : $price;
-      }
-    }
-
-    if (count($add_credit_totals)) {
-      /** @var \Apigee\Edge\Api\Monetization\Entity\SupportedCurrencyInterface[] $supported_currencies */
-      $supported_currencies = Drupal::service('apigee_m10n.monetization')
-        ->getSupportedCurrencies();
-
-      // Validate the total for each currency agains the minimum top up amount.
-      foreach ($add_credit_totals as $currency_code => $add_credit_total) {
-        if (isset($supported_currencies[$currency_code])) {
-          if (($supported_currency = $supported_currencies[$currency_code])
-            && ($minimum_top_up_amount = $supported_currency->getMinimumTopUpAmount())
-            && ((float) $add_credit_total->getNumber() < $minimum_top_up_amount)) {
-            $form_state->setErrorByName('review', t('The minimum top up amount for @currency_code is @amount.', [
-              '@currency_code' => $supported_currency->getName(),
-              '@amount' => Drupal::service('commerce_price.currency_formatter')->format($minimum_top_up_amount, $supported_currency->getName(), [
-                'currency_display' => 'code',
-              ]),
-            ]));
-          }
-        }
-      }
-    }
-  }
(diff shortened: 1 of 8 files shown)


modules/add_credit/apigee_m10n_add_credit.module
```diff
@@ -18,7 +18,7 @@
  * Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
  */
 
-use Drupal\commerce_product\Entity\ProductVariationInterface;
+use Drupal\commerce_checkout\Plugin\Commerce\CheckoutFlow\CheckoutFlowBase;
 use Drupal\Core\Entity\EntityInterface;
```
Jaesin, line 21: `CheckoutFlowBase` is unused

modules/add_credit/apigee_m10n_add_credit.module
```diff
@@ -51,6 +51,13 @@ function apigee_m10n_add_credit_entity_bundle_field_info(EntityTypeInterface $en
   return apigee_m10n_add_credit_service()->entityBundleFieldInfo($entity_type, $bundle, $base_field_definitions);
 }
 
+/**
+ * Implements hook_form_alter().
+ */
+function apigee_m10n_add_credit_form_alter(&$form, FormStateInterface $form_state, $form_id) {
+  apigee_m10n_add_credit_service()->formAlter($form, $form_state, $form_id);
+}
```
Jaesin, line 58: There is a precedent for using `formFormIdFormAlter` for form id specific alters. See: `\Drupal\apigee_m10n_add_credit\AddCreditServiceInterface::formCommerceProductTypeEditFormAlter` but this is not a blocker.

modules/add_credit/src/AddCreditService.php
```diff
@@ -19,7 +19,10 @@
 
 namespace Drupal\apigee_m10n_add_credit;
 
+use Drupal;
+use Drupal\apigee_m10n\MonetizationInterface;
```
Jaesin: Unused `MonetizationInterface`.

modules/add_credit/src/AddCreditService.php
```diff
@@ -257,6 +260,69 @@ public function fieldWidgetFormAlter(&$element, FormStateInterface $form_state,
...
+      $add_credit_totals = [];
+      foreach ($flow->getOrder()->getItems() as $item) {
+        /** @var \Drupal\commerce_product\Entity\ProductInterface $product */
+        if (($product = $item->getPurchasedEntity()->getProduct())
+          && ($product->hasField('apigee_add_credit_enabled'))
+          && ($product->get('apigee_add_credit_enabled')->value)) {
+          // Use the total price since this factors in the quantity.
+          $price = $item->getTotalPrice();
+          $currency_code = strtolower($price->getCurrencyCode());
+          $add_credit_totals[$currency_code] = !empty($add_credit_totals[$currency_code]) ? $add_credit_totals[$currency_code]->add($price) : $price;
```
Jaesin: I think this check makes since since someone could have an old cart and come back to check out after the minimum re-charge for a given currency was changed. I think we should enforce the minimum per line item since it's conceivable different line items could be for different accounts (or other unknown alterations).

modules/add_credit/tests/src/FunctionalJavascript/AddCreditCustomAmountTest.php
```diff
@@ -176,6 +167,95 @@ public function testUnitPriceValidation(float $minimum = NULL, float $maximum =
     $this->assertSession()->pageTextContains($message);
   }
 
+  /**
+   * Tests adding a custom amount and checking out.
+   */
+  public function testCustomAmountPriceCheckout() {
```
Jaesin, line 176: Oh, man  this test alone now takes 7 minutes on my local, Maybe we should add a todo to make this test more efficient. It looks like `@dataProvider` installs a copy of drupal for every each item.

modules/add_credit/src/AddCreditService.php
shadcn: @Jaesin yeah this handles multiple line items from different accounts too.

modules/add_credit/tests/src/FunctionalJavascript/AddCreditCustomAmountTest.php
shadcn: Hmm are you sure it's this test? This one does not use a `dataProvider`.

modules/add_credit/apigee_m10n_add_credit.module
shadcn: Fixed.

modules/add_credit/src/AddCreditService.php
shadcn: Fixed.

modules/add_credit/apigee_m10n_add_credit.module
shadcn: So this form has a dynamic form id. I used a form alter and then a check to see if we are on the review step.
```
if (($flow = $form_state->getFormObject())
      && ($flow instanceof CheckoutFlowBase)
      && ($form['#step_id'] == 'review')
    )
```

Jaesin: OK. That makes sense.

modules/add_credit/tests/src/FunctionalJavascript/AddCreditCustomAmountTest.php
Jaesin: I was referring to `AddCreditCustomAmountTest.php`.

modules/add_credit/src/AddCreditService.php
Jaesin: I see this adding all of the items with the the same total and then validating the total regardless of the `add_credit_account`.

Wouldn't it be easier to just make sure each line item is above the minimum amount.

shadcn: I think if you have 2 items that together makes the minimum amount, you should still be able to top up.

Jaesin: If you are setting up an add credit product, you should be making sure a one single item in the cart meets the minimum. We wanted to have validation on that when the product is saved regardless of if they are using the price range option but this solution at least prevents checking out with a top up that won't pass the check in the edge API. `\Drupal\apigee_m10n_add_credit\EventSubscriber\CommerceOrderTransitionSubscriber::handleOrderStateChange` does aggregate the amout before adding the credit to the account but that is just to reduce the number of calls to the API endpoint for one single order. The method also considers the `add_credit_account` data when calculating the totals.

We could try to consider all of the other aspects of the line item in this validaiton but the easiest solution is to validate per line item since the commerce product is supposed to have the minimum amount validated anyway.

shadcn: Done.

Jaesin (approved):
GTG

PR #54, opened 2019-02-26
[DRUP-596-597-599] add credit config, target selector and link for prepaid balances

shadcn:
This PR covers [DRUP-596], [DRUP-597] and [DRUP-599]. 

1. It adds a config page for add credit products.
2. It provides a base field of `add_credit_target_entity` type for the order item. This holds the reference to the developer / team entity and is use in `BalanceAdjustmentJob` to top up the corresponding account.
3. It adds a link on the prepaid balance pages to add credit to prepaid balances.
4. It adds new permissions for adding credit to any/own developer/team prepaid balances.
5. The `apigee_m10n_add_credit` module implements support for adding credit to the `developer` entity via `AddCreditEntityType` plugin.
6. The `apigee_m10n_teams` module provides implementation for `team`.

This depends on #52.

Some screenshots:

### DRUP-596
drup-596

### DRUP-597
drup-597

### DRUP-599
drup-599


modules/add_credit/src/Plugin/AddCreditEntityType/Developer.php
```diff
@@ -0,0 +1,75 @@
...
+use Drupal\apigee_edge\Entity\DeveloperInterface;
+use Drupal\apigee_m10n_add_credit\Annotation\AddCreditEntityType;
+use Drupal\apigee_m10n_add_credit\Plugin\AddCreditEntityTypeBase;
+use Drupal\Core\Entity\EntityInterface;
+use Drupal\Core\Session\AccountInterface;
+
+/**
+ * Defines a developer add credit entity type plugin.
+ *
+ * @AddCreditEntityType(
```
Jaesin, line 31: I think a plugin is overkill for this use case. We have 2 known types here to deal with and won't have any more so we don't need discovery.

modules/add_credit/src/Routing/AddCreditRoutes.php
```diff
@@ -0,0 +1,111 @@
...
+   *   An array of route objects.
+   */
+  public function routes() {
+    $routes = [];
+
+    foreach ($this->addCreditPluginManager->getPlugins() as $plugin) {
+      $entity_type_id = $plugin->getPluginId();
+      $path_entity_type_id = $plugin->getRouteEntityTypeId();
+      $routes["apigee_m10n_add_credit.add_credit.{$entity_type_id}"] = new Route(
+        $plugin->getPath(),
```
Jaesin: I don't think we need these routes, we should be able to link directly to the associated product.

modules/add_credit/src/Plugin/AddCreditEntityType/Developer.php
shadcn: I tried a non-plugin approach initially but there was a lot of module exists calls all over the place to handle teams, permissions, and the settable options for the field.

modules/add_credit/src/Routing/AddCreditRoutes.php
shadcn: OK I can make this change.

shadcn: This is completed.

modules/add_credit/src/AddCreditService.php
```diff
@@ -259,6 +275,90 @@ public function fieldWidgetFormAlter(&$element, FormStateInterface $form_state,
     }
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function apigeeM10nPrepaidBalancePageAlter(array &$build, EntityInterface $entity) {
+    // TODO: This can be move to entity operations when/if prepaid balance are
+    // made into entities.
+    if ((count($build['table']['#rows']))) {
```
Jaesin, line 297: We could easily fake operations with an operation hook instead of a page alter.

modules/add_credit/apigee_m10n_add_credit.links.task.yml
```diff
@@ -0,0 +1,14 @@
+apigee_m10n_add_credit.settings.add_credit:
+  route_name: apigee_m10n_add_credit.settings.add_credit
+  title: 'Add credit'
+  base_route: apigee_m10n.settings.prepaid_balance
+
+apigee_m10n_add_credit.settings.add_credit.products:
```
Jaesin: I wonder if we can combine these two settings forms.

modules/apigee_m10n_teams/src/Controller/TeamPrepaidBalanceController.php
```diff
@@ -26,6 +26,8 @@
 use Drupal\apigee_m10n_teams\Access\TeamPermissionAccessInterface;
 use Drupal\apigee_m10n_teams\Form\TeamPrepaidBalanceReportsDownloadForm;
 use Drupal\apigee_m10n_teams\TeamSdkControllerFactoryAwareTrait;
+use Drupal\Core\Entity\EntityInterface;
```
Jaesin, line 29: If we pull out this and `AddCreditEntityType/Team.php` we can compare this against 8.x-1.x.

src/ApigeeSdkControllerFactory.php
```diff
@@ -146,7 +146,7 @@ public function developerBalanceController(UserInterface $developer): DeveloperP
    * {@inheritdoc}
    */
   public function companyBalanceController(CompanyInterface $company): CompanyPrepaidBalanceControllerInterface {
-    $legal_name = $company->getLegalName();
+    $legal_name = $company->getName();
```
Jaesin: If we need to key the cache by name instead of legal name, we should also change the variable name.

modules/add_credit/apigee_m10n_add_credit.links.task.yml
Jaesin: With this PR, The add credit settings menu item is moved to a tab under the prepaid balance menu. Some of the settings are related to the prepaid balance and I guess adding credit is conceptually part of balance related but I'm not sure this menu structure is intuitive. We should look at this a little more.

Jaesin:
Another thing we need is version constraints on dependencies in `apigee_m10n_add_credit.info.yml`. Drupal commerce 2.9.0 produced some errors that blocked the selector from showing.

modules/add_credit/src/AddCreditService.php
shadcn: This isn't really a page alter. It's just named this way. Should probably rename this to `prepaid_balance_table_alter`.

shadcn: Renamed it to `prepaid_balance_list_alter` to be in line with other entity list.

modules/add_credit/apigee_m10n_add_credit.links.task.yml
shadcn: The products config is now part of the add credit config form.

Jaesin (approved):

PR #55, opened 2019-02-27
[DRUP-568] Teams TNC

minnur:
WIP - Depends on https://github.com/apigee/apigee-m10n-drupal/pull/47


modules/apigee_m10n_teams/src/TeamSdkControllerFactory.php
```diff
@@ -46,6 +47,17 @@ public function companyAcceptedRatePlanController(string $company_id): CompanyAc
     return $this->controllers[__FUNCTION__][$company_id];
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function companyTermsAndConditionsController(string $company_id): CompanyTermsAndConditionsController {
+    return new CompanyTermsAndConditionsController(
```
Jaesin: All of the other controllers are using static caching for these controlelrs. See: `\Drupal\apigee_m10n_teams\TeamSdkControllerFactory::companyAcceptedRatePlanController`.

modules/apigee_m10n_teams/src/MonetizationTeams.php
```diff
@@ -184,4 +240,133 @@ protected function teamAccessCheck(): TeamPermissionAccessInterface {
...
+      $this->companyAcceptedTermsStatus[$company_id] = ($latest instanceof LegalEntityTermsAndConditionsHistoryItem) && $latest->getAction() === 'ACCEPTED';
+    }
+
+    return $this->companyAcceptedTermsStatus[$company_id];
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getLatestTermsAndConditions(): ?TermsAndConditionsInterface {
```
Jaesin: This looks like a copy of `\Drupal\apigee_m10n\Monetization::getLatestTermsAndConditions(). We should inject the monetization service and use that to get the latest.

modules/apigee_m10n_teams/src/MonetizationTeams.php
```diff
@@ -184,4 +240,133 @@ protected function teamAccessCheck(): TeamPermissionAccessInterface {
...
+    return $this->latestTermsAndConditions;
+  }
+
+  /**
+   * Gets the full list of terms and conditions.
+   *
+   * @return \Apigee\Edge\Api\Monetization\Entity\TermsAndConditionsInterface[]
+   *   Returns the full list of terms and conditions or false on error.
+   */
+  protected function getTermsAndConditionsList(): array {
```
Jaesin: We  should also use the `apigee_m10n.monetization` service for this.

Jaesin:
I tested this manually and the team subscribe form was redirecting to the user subscribe form. Purchasing a plan for a team was never fully tested because of this missing piece.

Now that I fixed the URL redirect, the normal `apigee_tnc_widget` widget errors out because it expects a developer ID is present.

```
The website encountered an unexpected error. Please try again later.
TypeError: Argument 1 passed to Drupal\apigee_m10n\Monetization::acceptLatestTermsAndConditions() must be of the type string, null given, called in .../modules/contrib/apigee_m10n/src/Plugin/Field/FieldWidget/TermsAndConditionsWidget.php on line 150 in Drupal\apigee_m10n\Monetization->acceptLatestTermsAndConditions() (line 353 of modules/contrib/apigee_m10n/src/Monetization.php). 
```

minnur:
The issue should be fixed, I was able subscribe to a plan as a team.

PR #56, opened 2019-02-28
Redirect routes

Jaesin:
We discussed removing the leading `/user` from pathts that redirect to `/user/{user}/...`.


shadcn (approved):
Approved ✅

PR #58, opened 2019-03-01
[DRUP-554] No anonymous users.

Jaesin:
This prevents admins from allowing anonymous users access to monetization.

### DIFF
--- a/src/Monetization.php
+++ b/src/Monetization.php
@@ -426,10 +426,10 @@ class Monetization implements MonetizationInterface {
   public function formUserAdminPermissionsAlter(&$form, FormStateInterface $form_state, $form_id) {
     // Disable All incompatible permissions in the UI.
     foreach (array_keys($this->getMonetizationPermissions()) as $permission_name) {
-      if (isset($form['permissions'][$permission_name]['anonymous'])) {
+      if (isset($form['permissions'][$permission_name][AccountInterface::ANONYMOUS_ROLE])) {
         // Disable the permission.
-        $form['permissions'][$permission_name]['anonymous']['#disabled'] = TRUE;
-        $form['permissions'][$permission_name]['anonymous']['#value'] = 0;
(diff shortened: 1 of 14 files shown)


src/Monetization.php
```diff
@@ -406,4 +420,49 @@ public function getPrepaidBalanceReports(string $developer_id, \DateTimeImmutabl
...
+      }
+    }
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function userRolePresave(RoleInterface $user_role) {
+    // This prevents monetization permission grants on config import.
+    if ($user_role->id() === 'anonymous') {
```
shadcn: Suggestion: `AccountInterface::ANONYMOUS_ROLE`

src/Monetization.php
```diff
@@ -406,4 +420,49 @@ public function getPrepaidBalanceReports(string $developer_id, \DateTimeImmutabl
...
+   * Gets a list of Apigee monetization permissions.
+   *
+   * These permissions should only be applied to authenticated roles.
+   *
+   * @return array
+   *   Permissions for the monetization module.
+   */
+  protected function getMonetizationPermissions() {
+    return array_filter($this->permission_handler->getPermissions(), function ($permission) {
+      return ($permission['provider'] === 'apigee_m10n');
```
shadcn, line 464: Do we need to handle m10n submodules? I know we're going to have a few permissions in `apigee_m10n_add_credit`.

shadcn (commented):
@Jaesin I added a few comments.

src/Monetization.php
Jaesin: Yes, but those will be follow-ups to this.

src/Monetization.php
Jaesin: Good Idea

shadcn (approved):
Approved

PR #62, opened 2019-03-08
Devteams preparations.

Jaesin:
This PR contains all of changes to `apigee_m10n` and `apige_m10n_ad_credit` from the `8.x-1.x-devtrams` branch without the devteams work included.

### DIFF
--- a/apigee_m10n.permissions.yml
+++ b/apigee_m10n.permissions.yml
@@ -1,9 +1,6 @@
 administer apigee monetization:
   title: 'Administer apigee monetization'
 
-access monetization packages:
-  title: 'Access monetization packages'
-
 view mint prepaid reports:
   title: 'View Mint Prepaid Reports'
 
--- a/src/Entity/Controller/ContextDependentEntityController.php
+++ b/src/Entity/Controller/ContextDependentEntityController.php
@@ -46,7 +46,7 @@ class ContextDependentEntityController extends ControllerBase {
     // The developer routes should always follow this pattern.
     $developer_route_name = "entity.{$entity_type_id}.developer";
     // Add the current user as the develoepr (`user`).
-    $parameters = ['user' => \Drupal::currentUser()->id()];
+    $parameters = ['user' => $this->currentUser()->id()];
     // Re-add the rest of the parameters for this route.
     foreach ($route_match->getParameters()->keys() as $key) {
       $parameters[$key] = $route_match->getRawParameter($key);
--- a/src/Entity/Controller/ContextDependentEntityController.php
+++ b/src/Entity/Controller/ContextDependentEntityController.php
@@ -1,0 +1,58 @@
+<?php
+
+/*
+ * Copyright 2018 Google Inc.
+ *
+ * This program is free software; you can redistribute it and/or modify it under
+ * the terms of the GNU General Public License version 2 as published by the
+ * Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+ * or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public
+ * License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc., 51
+ * Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ */
+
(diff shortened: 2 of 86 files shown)


src/Entity/Controller/ContextDependentEntityController.php
```diff
@@ -0,0 +1,58 @@
...
+   * @return \Symfony\Component\HttpFoundation\RedirectResponse
+   *   A redirect response.
+   */
+  public function developerRedirect(RouteMatchInterface $route_match): RedirectResponse {
+    // Gets the entity type ID from the route options.
+    $entity_type_id = $route_match->getRouteObject()->getOption('entity_type_id');
+    // The developer routes should always follow this pattern.
+    $developer_route_name = "entity.{$entity_type_id}.developer";
+    // Add the current user as the develoepr (`user`).
+    $parameters = ['user' => \Drupal::currentUser()->id()];
```
shadcn: Inject this 👍

src/Entity/ListBuilder/SubscriptionListBuilderForDeveloper.php
```diff
@@ -19,309 +19,66 @@
...
    * {@inheritdoc}
    */
-  public static function create(ContainerInterface $container) {
-    $entityType = $container->get('entity_type.manager')->getDefinition('subscription');
-    return static::createInstance($container, $entityType);
-  }
+  public function render(UserInterface $user = NULL) {
+    // Return 404 if the user is not set and keep a compatible method signature.
+    if (!($user instanceof UserInterface)) {
+      return new RedirectResponse(Url::fromRoute('system.404'));
```
shadcn: Do we need a redirect here? What about an exception? `throw new NotFoundHttpException();`

shadcn: RedirectResponse requires a string: `return new RedirectResponse(Url::fromRoute('system.404')->toString());`

src/Entity/ParamConverter/SubscriptionConverter.php
```diff
@@ -37,39 +34,24 @@ class SubscriptionConverter extends EntityConverter implements ParamConverterInt
...
-    // If the entity type is translatable, ensure we return the proper
-    // translation object for the current context.
-    if ($entity instanceof EntityInterface && $entity instanceof TranslatableInterface) {
-      $entity = $this->entityManager->getTranslationFromContext($entity, NULL, ['operation' => 'entity_upcast']);
-    }
-
-    return $entity;
+    $user = (!$user || $user instanceof UserInterface) ? $user : User::load($user);
+    // Get the developer ID.
+    $develoepr_id = $user instanceof UserInterface ? $user->getEmail() : FALSE;
```
shadcn: s/develoepr_id/developer_id

src/Entity/Form/UnsubscribeConfirmForm.php
```diff
@@ -35,13 +35,6 @@
  */
 class UnsubscribeConfirmForm extends EntityConfirmFormBase {
```
shadcn, line 36: I'm seeing the following error when I try to unsubscribe:

`Error while cancelling plan: Property "keepOriginalStartDate" has not been initialized on instance of Apigee\Edge\Api\Monetization\Entity\StandardRatePlanRevision class. Expected type: "boolean".`

shadcn (changes requested):
Left some feedback.

apigee_m10n.permissions.yml
```diff
@@ -1,9 +1,6 @@
 administer apigee monetization:
   title: 'Administer apigee monetization'
 
-access monetization packages:
-  title: 'Access monetization packages'
-
 view mint prepaid reports:
   title: 'View Mint Prepaid Reports'
```
shadcn, line 5: We should probably create a ticket and rename this to "view prepaid balance". Not sure why this is using Mint!

Jaesin: There is a ticket to look at the permissions. [DRUP-555]

src/Entity/Form/UnsubscribeConfirmForm.php
Jaesin: I see it when trying to subscribe.

Jaesin: This will be fixed in a follow up task.

shadcn (approved):

PR #63, opened 2019-03-11
[DRUP-519] Add ability to cancel future plans.

minnur:


src/Entity/Form/UnsubscribeConfirmForm.php
```diff
@@ -159,10 +160,15 @@ public function buildEntity(array $form, FormStateInterface $form_state) {
     $values = $form_state->getValues();
     $end_type = $values['end_type'] ?? 'now';
 
-    $this->subscription->setEndDate($end_type == 'on_date'
-      ? new \DateTimeImmutable($values['endDate'])
-      : new \DateTimeImmutable('-1 day'));
-
+    // Cancel future date.
+    if ($end_type == 'now' && $this->subscription->getSubscriptionStatus() === SubscriptionInterface::STATUS_FUTURE) {
+      $this->subscription->setEndDate($this->subscription->getStartDate());
```
Jaesin, line 165: I see that setting the end date to the same day as the start date keeps it as future and gives it a canceled status. 
Do we need a new status for future canceled plans? They are stilled displayed as "Future".

I guess it wouldn't be too difficult to test and make sure that plan doesn't become active for that day.

Jaesin: This gets complicated because the time is being set on some end dates resulting in the `$start !== $end`.

Jaesin (approved):
Good to go

PR #66, opened 2019-03-12
Rename the add_credit module to apigee_m10n_add_credit

shadcn:
This simple PR renames the add_credit module to apigee_m10n_add_credit to match the other modules.


Jaesin (approved):
Only one directory renamed in this PR.

PR #72, opened 2019-03-18
Fix merge issue where the company ID patch gets re-added in `4ba84c4 `.

Jaesin:


minnur (approved):
PR looks good and I was able to verify the fix locally. Just make sure tests are running and should be ok to merge.

PR #77, opened 2019-03-23
[DRUP-642] The `keepOriginalStartDate` attribute is not always set on rate plan revisions.

minnur:
Apply https://github.com/apigee/apigee-client-php/issues/47 patch.


Jaesin (approved):
The patch allows us to subscribe to rate plan revisions that don't have `keepOriginalStartDate` set.

PR #82, opened 2019-04-03
[DRUP-689, DRUP-698] Add cache context to the packages page.

minnur:


Jaesin:
I'm unable to repro with the latest 8.x-1.x let's make sure this is still a problem first.

minnur:
@Jaesin I think it is still a problem. While logged in as admin I am unable to make purchase links to show correct user ID in the Purchase links here `user/[ID]/monetization/subscriptions` pages. The page would probably show your user ID the one you logged in with, but try changing it to other ID in order to accept a rate plan for that user. You will notice that the link still has the ID of the user that you logged in with.

Jaesin:
OK, I was able to validate the issue still exists. It's also happening on the package entity canonical route so I think the route user cache dependency needs to be at the entity level.

minnur:
> OK, I was able to validate the issue still exists. It's also happening on the package entity canonical route so I think the route user cache dependency needs to be at the entity level.

@Jaesin So where would that be, Apigee Edge module or entities in m10 module ?

Jaesin:
Fixed by: https://github.com/apigee/apigee-m10n-drupal/pull/97

PR #86, opened 2019-04-05
[DRUP-697] Purchased Plans page displays invalid data. Match output to Drupal 7 version

minnur:
This is now matching to what we see in Drupal 7. However it seems like we might need to review this more. There is a comment from @cnovak 
```
    // COMMERCE-558: If there is an end date and it has already started,
    // and the plan was also ended today, shift end_date to start_date.```

### DIFF
--- a/src/Entity/Subscription.php
+++ b/src/Entity/Subscription.php
@@ -64,7 +64,9 @@ use Drupal\user\UserInterface;
  */
 class Subscription extends FieldableEdgeEntityBase implements SubscriptionInterface, EntityOwnerInterface {
 
-  use EndDatePropertyAwareDecoratorTrait;
+  use EndDatePropertyAwareDecoratorTrait {
+    getEndDate as traitGetEndDate;
+  }
   use StartDatePropertyAwareDecoratorTrait;
 
   public const ENTITY_TYPE_ID = 'subscription';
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by Jaesin** [comment]
Hmm, I think this is failing because of the timezone. See: https://github.com/apigee/apigee-m10n-drupal/commit/a568b47da1c33542cba1fa791c6da8b13744f193

This was passing before. I guess we need to look at this again.

**Comment by Jaesin** [comment]
The comment mentioned above and the code from d7 didn't really match so I'm wondering which we want to do. Change the comment to match the code as in https://github.com/apigee/apigee-m10n-drupal/pull/86/commits/8eabf1413b583f8574a04c4a4be7eeb0c375ffd4 (`If the plan has ended, make sure the end date doesn't come before the start date`) or change the code to match the comment. That would look more like this:

```
  /**
   * Adjust the date so it never ends before it starts.
   *
   * @return \DateTimeImmutable|null
   *   The adjusted end date.
   *
   * @throws \Exception
   */
  protected function getAdjustedEndDate() : ?\DateTimeImmutable {
    $today = new \DateTime('today', $this->getRatePlan()->getOrganization()->getTimezone());

    // COMMERCE-558: If there is an end date and it has already started,
    // and the plan was also ended today, shift end_date to start_date.
    if (($end_date = $this->traitGetEndDate())
      && $end_date->format('Y-m-d') === $today->format('Y-m-d')
      && $end_date < $this->getStartDate()
    ) {
      return $this->getStartDate();
    }
    else {
      return $end_date;
    }
  }
```

Jaesin (approved):
This is approved for the date adjustment.

PR #87, opened 2019-04-07
Fix redirect issue on accept and cancel rate plan actions.

minnur:


src/Entity/Subscription.php
```diff
@@ -336,6 +336,18 @@ public function getDeveloper(): ?DeveloperInterface {
     return $this->decorated->getDeveloper();
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function getDrupalDeveloperId(): int {
+    $users = \Drupal::entityTypeManager()->getStorage('user')
```
Jaesin: This looks very familiar. =)  See: https://github.com/apigee/apigee-m10n-drupal/pull/85/commits/24f17c300bf946db38a1c2cd09a4c7bb73411a1e

You could cherry pick `24f17c3` but it might be best to wait until the PR for 555 is merged and then use `$this->entity->getOwnerId()` where the UID is needed. Only because implmenting `EntityOwnerInterface` with the current handlers might have some unexpected effects.

minnur:
I am going to re-implement this when https://github.com/apigee/apigee-m10n-drupal/pull/85 merged

Jaesin (approved):
Rebased to fix merge conflict. Manually tested.

PR #93, opened 2019-04-22
[DRUP-699] Insufficient funds workflow

arlina-espinoza:
On this PR:

- If trying to subscribe to a plan and user does not have enough funds, it will disable the "purchase" button, and display a message _"You have insufficient funds to purchase plan <plan>"_, and a link to add credit in the appropriate currency.
- However, calculating the _minimum balance needed_ is tricky, as there are multiple types of plans, prepaid/advance, with/without fees, freemium, etc, and the SDK doesn't specify this minimum upfront, so the only calculation currently done to display the message is checking that the current balance covers at least the "setup fee", if one is specified. (See https://docs.apigee.com/api-platform/monetization/create-rate-plans.html#rateplanops).
- For other cases where more credit is needed, the user will be shown a user friendly error message when trying to purchase a plan with a link to the appropriate "add credit" product.

### DIFF
--- a/modules/apigee_m10n_add_credit/apigee_m10n_add_credit.module
+++ b/modules/apigee_m10n_add_credit/apigee_m10n_add_credit.module
@@ -114,9 +114,5 @@ function apigee_m10n_add_credit_commerce_product_access(EntityInterface $entity,
  *   The `apigee_m10n.add_credit` service.
  */
 function apigee_m10n_add_credit_service() {
-  // Use static caching for performance.
-  static $service;
-  $service = $service ?: \Drupal::service('apigee_m10n.add_credit');
-
-  return $service;
+  return \Drupal::service('apigee_m10n.add_credit');
 }
--- a/modules/apigee_m10n_add_credit/apigee_m10n_add_credit.module
+++ b/modules/apigee_m10n_add_credit/apigee_m10n_add_credit.module
@@ -108,13 +108,6 @@ function apigee_m10n_add_credit_commerce_product_access(EntityInterface $entity,
 }
 
 /**
- * Implements hook_form_FORM_ID_alter().
- */
-function apigee_m10n_add_credit_form_subscription_form_alter(&$form, FormStateInterface $form_state, $form_id) {
-  apigee_m10n_add_credit_service()->subscriptionFormAlter($form, $form_state, $form_id);
-}
-
-/**
  * Helper to get the `apigee_m10n.add_credit` service.
  *
  * @return \Drupal\apigee_m10n_add_credit\AddCreditServiceInterface
--- a/modules/apigee_m10n_add_credit/apigee_m10n_add_credit.module
+++ b/modules/apigee_m10n_add_credit/apigee_m10n_add_credit.module
@@ -18,10 +18,12 @@
  * Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
  */
 
+use Drupal\apigee_m10n\Entity\SubscriptionInterface;
 use Drupal\Core\Entity\EntityInterface;
 use Drupal\Core\Entity\EntityTypeInterface;
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Session\AccountInterface;
+use Drupal\Core\StringTranslation\TranslatableMarkup;
 
 /**
  * Implements hook_mail().
@@ -105,6 +107,13 @@ function apigee_m10n_add_credit_apigee_m10n_prepaid_balance_list_alter(array &$b
  */
 function apigee_m10n_add_credit_commerce_product_access(EntityInterface $entity, $operation, AccountInterface $account) {
   return apigee_m10n_add_credit_service()->commerceProductAccess($entity, $operation, $account);
+}
+
+/**
+ * Implements hook_apigee_m10n_insufficient_balance_error_message_alter().
+ */
+function apigee_m10n_add_credit_apigee_m10n_insufficient_balance_error_message_alter(TranslatableMarkup &$message, SubscriptionInterface $subscription) {
+  return apigee_m10n_add_credit_service()->insufficientBalanceErrorMessageAlter($message, $subscription);
 }
 
 /**
--- a/modules/apigee_m10n_add_credit/apigee_m10n_add_credit.module
+++ b/modules/apigee_m10n_add_credit/apigee_m10n_add_credit.module
@@ -108,6 +108,13 @@ function apigee_m10n_add_credit_commerce_product_access(EntityInterface $entity,
 }
 
 /**
+ * Implements hook_form_FORM_ID_alter().
+ */
+function apigee_m10n_add_credit_form_subscription_form_alter(&$form, FormStateInterface $form_state, $form_id) {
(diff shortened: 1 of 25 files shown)


arlina-espinoza:
Added an admin option to optionally disable the insufficient funds workflow pre-check in the module config page _(/admin/config/apigee-edge/monetization/add-credit)_.

modules/apigee_m10n_add_credit/src/AddCreditService.php
```diff
@@ -503,4 +525,61 @@ public static function inlineEntityFormTableFieldsPriceCallback(ProductVariation
     return t('N/A');
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function subscriptionFormAlter(&$form, FormStateInterface $form_state, $form_id) {
```
Jaesin: I'm wondering if this should be a part of the `apigee_m10n` module. Some monetization dev portals may not be using the `apigee_m10n_add_credit` module because they are using recurly or some other custom solution. In that case they would loose validation.

arlina-espinoza: I've moved the logic to the main _apigee_m10n_ module. The configuration option to disable the workflow is now on the "Prepaid balance" config page (_/admin/config/apigee-edge/monetization/prepaid-balance_).
Screenshot_2019-04-26 Prepaid balance

arlina-espinoza:
Updated the code so the validation will only be displayed if the developer type is prepaid.

modules/apigee_m10n_add_credit/src/AddCreditService.php
cnovak: If a developer is Prepaid, insufficient balance checks should never be done since they should not see a prepaid balance tab.

src/Entity/Form/SubscriptionForm.php
```diff
@@ -125,7 +169,97 @@ public function save(array $form, FormStateInterface $form_state) {
       }
     }
     catch (\Exception $e) {
-      $this->messenger->addError($e->getMessage());
+      $previous = $e->getPrevious();
+
+      // If insufficient funds error, format nicely and add link to add credit.
+      if ($previous instanceof ClientErrorException && $previous->getEdgeErrorCode() === static::INSUFFICIENT_FUNDS_ERROR) {
+        preg_match_all('/\[(?\'amount\'.+)\]/', $e->getMessage(), $matches);
```
cnovak, line 224: This looks like something the client php library should do, handle and extract errors into clean objects along with tests to validate the errors clients depend on are not breaking.

Jaesin:
I rebased against 8.x-1.x. It still needs tests.

src/MonetizationInterface.php
```diff
@@ -218,4 +218,21 @@ public function formUserAdminPermissionsAlter(&$form, FormStateInterface $form_s
...
+   *
+   * @param string $currency_id
+   *   The currency ID.
+   * @param \Drupal\user\UserInterface $account
+   *   The user entity for which to generate the fallback link.
+   *
+   * @return \Drupal\Core\Url
+   *   The URL object.
+   */
+  public function getAddCreditUrl($currency_id, UserInterface $account);
```
Jaesin: I'm not so keen on this one. I'm thinking the add credit module should be the one to alter to add information only it is aware of.

cnovak:
Both @Jaesin and @arlina-espinoza have signed CLA

cnovak:
@arlina-espinoza @Jaesin both signed CLAs

src/Entity/Form/SubscriptionForm.php
```diff
@@ -156,13 +211,98 @@ public function save(array $form, FormStateInterface $form_state) {
...
+        $rate_plan = $this->getEntity()->getRatePlan();
+        $currency_id = $rate_plan->getCurrency()->id();
+
+        $amount_formatted = $this->monetization->formatCurrency($matches['amount'][0], $currency_id);
+        $insufficient_funds_error_message = $this->t('You have insufficient funds to purchase plan %plan. @adenndum', [
+          '%plan' => $rate_plan->label(),
+          '%amount' => $amount_formatted,
+          '@adenndum' => $amount ? $this->t('To purchase this plan you are required to add at least %amount to your account.', ['%amount' => $amount_formatted]) : '',
+        ]);
+        $this->module_handler->alter('apigee_m10n_insufficient_balance_error_message', $insufficient_funds_error_message, $this->getEntity());
```
cnovak: I am getting  an error at this line with apigee_m10n_add_credit module *not* enabled:
```
Notice: Only variables should be passed by reference in Drupal\apigee_m10n\Entity\Form\SubscriptionForm->save() (line 235 of modules/contrib/apigee_m10n/src/Entity/Form/SubscriptionForm.php).
Drupal\apigee_m10n\Entity\Form\SubscriptionForm->save(Array, Object)
```

```suggestion
        $subscription_entity = $this->getEntity();
        $this->module_handler->alter('apigee_m10n_insufficient_balance_error_message', $insufficient_funds_error_message, $subscription_entity);
```

Jaesin: Fixed via git.

modules/apigee_m10n_add_credit/apigee_m10n_add_credit.module
```diff
@@ -107,16 +109,19 @@ function apigee_m10n_add_credit_commerce_product_access(EntityInterface $entity,
...
+  return apigee_m10n_add_credit_service()->insufficientBalanceErrorMessageAlter($message, $subscription);
+}
+
 /**
  * Helper to get the `apigee_m10n.add_credit` service.
  *
  * @return \Drupal\apigee_m10n_add_credit\AddCreditServiceInterface
  *   The `apigee_m10n.add_credit` service.
  */
 function apigee_m10n_add_credit_service() {
```
shadcn, line 125: This helper is not doing much now. In a future PR, we could call `\Drupal::service('apigee_m10n.add_credit')` directly.

src/Entity/Form/SubscriptionForm.php
```diff
@@ -52,17 +61,47 @@ class SubscriptionForm extends FieldableMonetizationEntityForm {
...
+   * @var \Drupal\Core\Config\ConfigFactoryInterface
+   */
+  protected $config;
+
+  /**
+   * The module handler.
+   *
+   * @var \Drupal\Core\Extension\ModuleHandlerInterface
+   */
+  protected $module_handler;
```
shadcn, line 83: `ModuleHandler`

src/Form/PrepaidBalanceConfigForm.php
```diff
@@ -103,6 +103,21 @@ public function buildForm(array $form, FormStateInterface $form_state) {
       '#description' => $this->t('Set the cache age for the prepaid balance for a developer.'),
     ];
 
+    $form['general'] = [
+      '#type' => 'fieldset',
+      '#title' => $this->t('General'),
+    ];
+
+    $description = 'If checked, the "Purchase" button on subscriptions will be
```
shadcn, line 111: We could inline this.

shadcn (approved):
I tested this. Everything is working as expected. I added a few minor comments below.

DRUP-699

PR #100, opened 2019-05-10
[DRUP-725] Remove apigee-client-php patch

shadcn:
Now that we have a new release for apigee-client-php and apigee-edge, we can remove the patch for it. https://github.com/apigee/apigee-client-php/releases/tag/2.0.2

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,12 +24,5 @@
         "sort-packages": true
     },
     "minimum-stability": "dev",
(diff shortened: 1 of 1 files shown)


cnovak (approved):
Lgtm

Jaesin:
`composer.json` should require `"apigee/apigee-client-php": "~2.0.2"` in order to drop the patch.

PR #101, opened 2019-05-15
[DRUP-702] Handle plan conflicts

shadcn:
Taking over this ticket from @minnur. Let's see what breaks here. Go CI!

<hr />

This is now ready for review.

Purchase_Test_3___Drupal_8_m10n-2

Purchase_Test_3___Drupal_8_m10n

### DIFF
--- a/modules/apigee_m10n_teams/src/Entity/Form/TeamSubscriptionForm.php
+++ b/modules/apigee_m10n_teams/src/Entity/Form/TeamSubscriptionForm.php
@@ -33,13 +33,6 @@ class TeamSubscriptionForm extends SubscriptionForm {
   /**
    * {@inheritdoc}
    */
-  public function form(array $form, FormStateInterface $form_state) {
-    return parent::form($form, $form_state);
-  }
-
-  /**
-   * {@inheritdoc}
-   */
   public function save(array $form, FormStateInterface $form_state) {
     try {
       if ($this->entity instanceof TeamsSubscriptionInterface && $this->entity->isTeamSubscription()) {
--- a/modules/apigee_m10n_teams/src/Entity/Form/TeamSubscriptionForm.php
+++ b/modules/apigee_m10n_teams/src/Entity/Form/TeamSubscriptionForm.php
@@ -33,6 +33,13 @@ class TeamSubscriptionForm extends SubscriptionForm {
   /**
    * {@inheritdoc}
    */
+  public function form(array $form, FormStateInterface $form_state) {
+    return parent::form($form, $form_state);
+  }
+
+  /**
+   * {@inheritdoc}
+   */
   public function save(array $form, FormStateInterface $form_state) {
     try {
       if ($this->entity instanceof TeamsSubscriptionInterface && $this->entity->isTeamSubscription()) {
--- a/src/Entity/Form/SubscriptionForm.php
+++ b/src/Entity/Form/SubscriptionForm.php
@@ -20,6 +20,7 @@
 namespace Drupal\apigee_m10n\Entity\Form;
 
 use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\Routing\CurrentRouteMatch;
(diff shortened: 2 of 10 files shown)


modules/apigee_m10n_teams/src/Entity/Form/TeamSubscriptionForm.php
```diff
@@ -30,6 +30,13 @@
  */
 class TeamSubscriptionForm extends SubscriptionForm {
 
+  /**
+   * {@inheritdoc}
+   */
+  public function form(array $form, FormStateInterface $form_state) {
```
Jaesin: This can be omitted since it doesn't do anything.

src/Entity/Form/SubscriptionForm.php
```diff
@@ -91,6 +91,15 @@ protected function actions(array $form, FormStateInterface $form_state) {
     // Set the save label if one has been passed into storage.
     if (!empty($actions['submit']) && ($save_label = $form_state->get('save_label'))) {
       $actions['submit']['#value'] = $save_label;
+      $actions['submit']['#button_type'] = 'primary';
+      if ($items = $form_state->get('planConflicts')) {
+        $parameters = \Drupal::routeMatch()->getParameters()->all();
```
Jaesin: You can inject `current_route_match`.

src/Entity/Form/SubscriptionForm.php
```diff
@@ -91,6 +91,15 @@ protected function actions(array $form, FormStateInterface $form_state) {
     // Set the save label if one has been passed into storage.
     if (!empty($actions['submit']) && ($save_label = $form_state->get('save_label'))) {
       $actions['submit']['#value'] = $save_label;
+      $actions['submit']['#button_type'] = 'primary';
+      if ($items = $form_state->get('planConflicts')) {
```
Jaesin: This would only be set after an unsuccessful subscribe action. Are we meant to be pro-active about knowing if about the conflicting plans and providing that plans will be overwritten?

shadcn: @Jaesin This is how it works on the d7 version. AFAIK we haven't talked about changing it. But I can see it being useful to have it on the form right away.

Jaesin:
When the teams module is enabled, I get the following error: 

```
The website encountered an unexpected error. Please try again later.
TypeError: Return value of Drupal\apigee_m10n\Entity\Subscription::getSuppressWarning() must be of the type boolean, null returned in Drupal\apigee_m10n\Entity\Subscription->getSuppressWarning() (line 364 of modules/contrib/apigee_m10n/src/Entity/Subscription.php). 
```

I think `$this->suppressWarning = FALSE;` needs to be added to `\Drupal\apigee_m10n_teams\Entity\TeamsSubscription::__construct` or `getSuppressWarning` should cast to a boolean value.

src/Entity/Form/SubscriptionForm.php
```diff
@@ -126,7 +156,116 @@ public function save(array $form, FormStateInterface $form_state) {
     }
     catch (\Exception $e) {
       $this->messenger->addError($e->getMessage());
+      $messages = $this->messenger->all();
+      $this->messenger->deleteAll();
+      if (!empty($messages['error'][0])) {
+        $overlap_error = $messages['error'][0];
+        // Make sure this is overlapping error message.
+        if (strstr($overlap_error, 'has following overlapping')) {
```
Jaesin: Shouldn't we be checking the client error with something like:

```
if (($client_error = $e->getPrevious())
  && $client_error instanceof ClientErrorException
  && $client_error->getEdgeErrorCode() === 'mint.developerHasFollowingOverlapRatePlans'
) {
```

shadcn: I guess Minnur found this from the legacy code. This is how they implemented that.

This is much better. Nice find. I'll implement this.

Jaesin (approved):
Awesome. Nice work.

PR #112, opened 2019-06-03
The rate plan converter failed to throw a 404 when the rate plan is n…

Jaesin:
…ot loaded.


shadcn (approved):
Looks good. This successfully returns a 404 for non-existing rate plans. Thanks @Jaesin

shadcn:
@Jaesin LGTM.

Issue #113, opened 2019-06-03
Provide Default Config and Ensure it's Possible to Customize Rate Plans According to Designs in M10n module


cnovak:
This task should also cover anything that is needed at field level to deal with volume bands, basic and future rate plans, etc.

## Relevant comps

Basic Rate plan: https://projects.invisionapp.com/d/main/default/#/console/16939246/360880462/preview

Rate plan with current and future in place:
https://projects.invisionapp.com/d/main/default/#/console/16939246/360880463/preview

PR #122, opened 2019-06-07
[Fixes #117] Display plans instead of packages on the pricing and plans page.

### LINKED ISSUE #117: Pricing & Plans: Implement According to New Designs Using Rate Plans
* The page shows Rate plans instead of Packages.
* Should probably use the teaser view mode (in m10n)
* Shows the Title, Description, Products, and View Pricing link

Design comps: https://projects.invisionapp.com/d/main/default/#/console/16939246/359615112/preview

Jaesin:


Jaesin:
Failure because of: https://www.drupal.org/node/3035980

src/Controller/PricingAndPlansController.php
```diff
@@ -107,63 +130,53 @@ public function catalogPage(UserInterface $user) {
...
-        $product_labels[] = $item->entity->label();
-      }
-      $build[$index]['#prefix'] = new FormattableMarkup('<li class="apigee-sdk-package">' . $header_html, [
-        '@package_label' => $build[$index]['#package']->label(),
-        '@product_labels' => implode(', ', $product_labels),
-      ]);
-      $build[$index]['#suffix'] = '</div></li>';
+  protected function buildPage($plans) {
+    $build = [
+      '#type' => 'container',
```
jacine: @Jaesin Can you please add this line? We need to be able to target that container.

`'#theme' => ['container__rate_plan_list'],`

Thanks!

EDIT:  Don't use this, see below.

jacine: @Jaesin Actually, apparently `#theme_wrappers` is better in this case, because of processing that happens.  Using `#theme` apparently results in [another rendering step](https://git.drupalcode.org/project/drupal/blob/8.8.x/core/lib/Drupal/Core/Render/RendererInterface.php#L204-219), and`{{ children }}` (which is what the default template implementation uses) no longer works because it's already rendered as #markup and lives under `{{ element }}`.  Using `#theme_wrappers` doesn't change that pipeline and gives the ability to customize as we need:

`'#theme_wrappers' => ['container__rate_plan_list'],`

src/Controller/PricingAndPlansController.php
```diff
@@ -107,63 +130,53 @@ public function catalogPage(UserInterface $user) {
...
+
+    foreach ($plans as $plan) {
+      // TODO: Add a test for render cache.
+      $build['#cache']['tags'] = Cache::mergeTags($build['#cache']['tags'], $plan->getCacheTags());
     }
+    // Get the view mode from package config.
+    $view_mode = $this->config(RatePlanConfigForm::CONFIG_NAME)->get('catalog_view_mode');
+
+    // Generate a build array using the view builder.
+    $build['plan_list'] = $this->entityTypeManager()->getViewBuilder('rate_plan')->viewMultiple($plans, $view_mode ?? 'default');
```
jacine: Can you please make this an item-list?  In order to control the layout of this I need to be able wrap these in column classes, and doing that on the entity view mode itself is a bad practice that may break stuff elsewhere for people in the future.

jacine: BTW, will need `'#theme' => ['item_list__rate_plan_list'],` for this, so it can be targeted as well.

Jaesin: Using automatic theme suggestions like `'#theme' => 'item_list__rate_plan_list'` or `'#theme' => 'container__rate_plan_list'` is a great idea but item list doesn't play well with the entity list builder because the entity list builder adds `#pre_render` and `#sorted` in which the former crashes the item-list template.

jacine: What do you suggest I do for wrapping these in columns?   Can you make a theme function that allows me to wrap each row?

Jaesin: We switched from using `view_builder->viewMultiple()` to `view_builder->view()` for each rate plan to make the list more compatible with containers/lists.

config/install/apigee_m10n.rate_plan.config.yml
```diff
@@ -0,0 +1 @@
+catalog_view_mode: "teaser"
```
jacine, line 1: 👍

config/install/core.entity_view_display.rate_plan.rate_plan.teaser.yml
```diff
@@ -0,0 +1,70 @@
...
+content:
+  description:
+    type: string
+    weight: 1
+    region: content
+    label: visually_hidden
+    settings:
+      link_to_entity: false
+    third_party_settings: {  }
+  displayName:
```
jacine, line 21: I think we should hide this one.  We can do it in the next round, since it's really more specific to the rate plan work and not this plans page.

css/apigee_m10n.plan-list.css
```diff
@@ -0,0 +1,27 @@
+.pricing-and-plans > .pricing-and-plans__item > .apigee-package-rate-plan {
```
jacine, line 1: This file needs work. But again, we'll do that in the next round and I'll help.

apigee_m10n.libraries.yml
```diff
@@ -1,5 +1,4 @@
 package:
-  version: 1.0
```
jacine, line 2: 👍

jacine (approved):
Some follow ups, but they should probably be handled in #113, and I want to help with those, so let's get this in. :)

Issue #111, opened 2019-06-03
"Configuration" admin menu broken after install


cnovak:
## Description
If you install the Apigee Edge module and the M10n module (and no Admin Toolbar module), you cannot configure Edge once installing the modules.

## Steps to reproduce
1. Install D8 using drupal composer: `composer create-project drupal-composer/drupal-project:8.x-dev d8-devportal --stability dev --no-interaction`
2. After site is installed, install Apigee Edge and Apigee Monetization ( issues do not happen when you select only Apigee edge)
3. Click on "Configuration", in order to navigate to configure Edge connection settings.

__Expected:__ You will see a listing of sub menu items that are under the “Configuration” menu item
__Actual:__ The “cannot reach management server” error is shown, with no way to reach the Edge connection settings page to fix the issue.

<img width="1331" alt="Screen Shot 2019-06-03 at 1 41 12 PM" src="https://user-images.githubusercontent.com/613093/58829734-ca63b000-8605-11e9-923d-b630d83c7fc2.png">

shadcn:
Fixed: https://github.com/apigee/apigee-m10n-drupal/pull/124

PR #124, opened 2019-06-11
[#111] Fix issue with requirements check on system configuration page

shadcn:
Issue: https://github.com/apigee/apigee-m10n-drupal/issues/111

### DIFF
--- a/apigee_m10n.install
+++ b/apigee_m10n.install
@@ -18,6 +18,7 @@
  * Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
  */
 
+use Apigee\Edge\Exception\ApiException;
 use Drupal\apigee_m10n\Monetization;
 use Drupal\apigee_m10n\MonetizationInterface;
 use Drupal\user\RoleInterface;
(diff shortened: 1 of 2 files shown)


shadcn:
❌ Failure: https://circleci.com/gh/arshad/apigee-m10n-drupal/663
✅ Fixed: https://circleci.com/gh/arshad/apigee-m10n-drupal/668

apigee_m10n.install
```diff
@@ -35,15 +35,20 @@ function apigee_m10n_requirements($phase) {
...
+      if (!$monetization->isMonetizationEnabled()) {
+        $requirements['apigee_monetization_connection_error'] = [
+          'title' => t('Apigee Monetization'),
+          'description' => t(Monetization::MONETIZATION_DISABLED_ERROR_MESSAGE),
+          'severity' => REQUIREMENT_ERROR,
+        ];
+      }
+    }
+    catch (Exception $exception) {
+
```
kscheirer: should probably do something with the exception, maybe log to watchdog?

Jaesin: I don't think we want to just hide all exceptions here. We should probably be catching `\Apigee\Edge\Exception\ApiException` here.

shadcn: Done and done.

Jaesin (approved):
This has been testing as is working.

PR #128, opened 2019-06-12
[Fixes #125] Change the subscription entity label to purchased plan.

### LINKED ISSUE #125: Change user facing strings of "Subscription/s" to "Purchase Plan/s"
As part of this task the following things should be changed:

1. The subscription entity's annotation labels should be changed to "Purchased plan/s".

2. The table headers on `/user/1/monetization/subscriptions` don't use the entity label):
  - From "Active and Future Subscriptions" to "Active and Future Plans"
  - From "Cancelled and Expired Subscriptions" to  "Canceled and Expired Plans"

3. Change the empty messages on `/user/1/monetization/subscriptions` to be a single generic and not using the entity label, e.g. "There are no results to display".

Jaesin:
Fixes #125

### DIFF
--- a/src/Entity/Subscription.php
+++ b/src/Entity/Subscription.php
@@ -35,11 +35,14 @@ use Drupal\user\UserInterface;
 /**
  * Defines the Subscription entity class.
  *
+ * The label was changed to purchased plan after the entity was created because
+ * it was decided it was a better match.
+ *
  * @\Drupal\apigee_edge\Annotation\EdgeEntityType(
  *   id             = "subscription",
- *   label          = @Translation("Subscription"),
- *   label_singular = @Translation("Subscription"),
- *   label_plural   = @Translation("Subscriptions"),
+ *   label          = @Translation("Purchased Plan"),
+ *   label_singular = @Translation("Purchased Plan"),
+ *   label_plural   = @Translation("Purchased Plans"),
  *   label_count = @PluralTranslation(
  *     singular = "@count Subscription",
  *     plural   = "@count Subscriptions",


src/Entity/Subscription.php
```diff
@@ -35,11 +35,14 @@
...
  * @\Drupal\apigee_edge\Annotation\EdgeEntityType(
  *   id             = "subscription",
- *   label          = @Translation("Subscription"),
- *   label_singular = @Translation("Subscription"),
- *   label_plural   = @Translation("Subscriptions"),
+ *   label          = @Translation("Purchased Plan"),
+ *   label_singular = @Translation("Purchased Plan"),
+ *   label_plural   = @Translation("Purchased Plans"),
  *   label_count = @PluralTranslation(
  *     singular = "@count Subscription",
```
jacine, line 47: What about the rest?

Jaesin: It looked like the other points of #125 were called into question. @cnovak Do we want to use what we have now which results in "Active and Future Purchased Plans", "Cancelled and Expired Purchased Plans", "No active or future purchased plans." and "No cancelled or expired purchased plans." or do we want to switch to hard coded values?

jacine: Yeah, sorry.... I am utterly confused about the convo we had.  So, best to have @cnovak look at it.

Jaesin:
This has been applied as part of https://github.com/apigee/apigee-m10n-drupal/pull/134

src/Entity/Subscription.php
cnovak: Looks good, the purchased plans is now using the entity labels along with the empty list message.  I think you may be concerned that it says "purchased plans" instead of just "plans".  We could modify it, but let's see if we get any feedback before changing.

jacine: Sort of... I was wondering about the lines directly below these (46+) that remained as "Subscription".  But in #134, [they were all changed](https://github.com/apigee/apigee-m10n-drupal/blob/8.x-1.x/src/Entity/PurchasedPlan.php#L46-L48), so this comment of mine is moot.   Thanks for chiming in here to help clarify! :smile:

PR #134, opened 2019-06-19
[M10N-126] Rename subscription entity.

### LINKED ISSUE #126: Change Subscription entity to PurchasedPlans
Change `Subscription` entity to `PurchasedPlan`.  The labels should be already changed to "purchased plan" and "purchased plans" via #125, but this ticket is to rename the actual entity to make it match our decision to call this a purchased plan instead of a subscription. 

As discussed yesterday, this would be good to do before 1.0. 

This relates to #125, but is substantially more work, which is why it's in a separate issue.

## Background
The Edge API there is a `plan` entity, but after you purchase a `plan` you can make calls to the `v1/mint/organizations/:org_name/developers/:developer_id/developer-accepted-rateplans` to get a list of entities which we are naming `PurchasedPlans`.

Jaesin:
Fixes #126.


shadcn (approved):
This all looks good. Reviewed and tested. Everything is working as expected. 💯  @Jaesin 

(Just a quick note for anyone else testing this. The purchase plan entity loses the "Purchase" link. You will need to add it back in the view displays).

LGTM @Jaesin 👍

Jaesin:
The "Purchase" link should be configured properly when you re-install.

PR #137, opened 2019-06-21
[Fixes #113] Allow rate plans to be configured for new design concept.

### LINKED ISSUE #113: Provide Default Config and Ensure it's Possible to Customize Rate Plans According to Designs in M10n module

Jaesin:
Adds a future start date and future plan links rate plan fields.


apigee_m10n.libraries.yml
```diff
@@ -14,6 +14,16 @@ rate_plan.entity_list:
...
+rate_plan.future_links_field:
+  css:
+    theme:
+      css/apigee_m10n.future-links-field.css: {}
+
+rate_plan.details_field:
+  css:
+    theme:
+      css/apigee_m10n.plan-details-field.css: {}
+
```
jacine, line 26: I don't see the point of making 2 separate libraries.  Why not just have one `rate_plan.details`?

apigee_m10n.module
```diff
@@ -52,9 +52,35 @@ function apigee_m10n_theme($existing, $type, $theme, $path) {
...
+      'base hook' => 'rate_plan_detail',
+    ],
+    'rate_plan_detail__ratecard' => [
+      'base hook' => 'rate_plan_detail',
+    ],
+    'rate_plan_detail__revshare_ratecard' => [
+      'base hook' => 'rate_plan_detail',
+    ],
+    'rate_plan_detail__adjustable_notification' => [
+      'base hook' => 'rate_plan_detail',
```
jacine: All four of these, and the templates associated should be removed.  The only reason you'd need to do this is if the templates had actual differences.  There are no differences, and keeping this would mean anyone trying to override the base template would need to override all 4 of these, plus the base.

apigee_m10n.module
```diff
@@ -75,3 +101,13 @@ function apigee_m10n_form_user_admin_permissions_alter(&$form, FormStateInterfac
...
   \Drupal::service('apigee_m10n.monetization')->userRolePresave($entity);
 }
+
+/**
+ * Implements hook_preprocess_field().
+ */
+function apigee_m10n_preprocess_field(&$variables) {
+  // Add a library to the rate plan `futurePlanLinks` field.
+  if ($variables['field_name'] === 'futurePlanLinks') {
+    $variables['#attached']['library'][] = 'apigee_m10n/rate_plan.future_links_field';
```
jacine, line 101: This should be attached via the render array in `viewElements()`, instead of preprocess.

templates/rate-plan-detail.html.twig
```diff
@@ -24,44 +29,61 @@
...
+      <div class="field field--inline">
+        <div class="field__label">{{ detail_field.label }} </div>
+        <div>{{ detail_field.value }} {{ (detail_field.suffix) ? '%' }}</div>
+      </div>
+    {% endif %}
+  {% endfor %}
+</div>
+
+{% if (ratecard_rates) %}
+  {% include 'ratecard-rates.html.twig' with {'rates': ratecard_rates} %}
```
jacine: We should not be using Twig include here for this.  This should just be able to render.  Also, don't need parenthesis on line 83.

templates/rate-plan-detail.html.twig
```diff
@@ -24,44 +29,61 @@
...
-<table class="rate-plan-detail">
-{% for detail_field in detail_fields %}
-  {% if detail_field.value %}
-  <tr class="field--label-inline">
-    <td><label class="field__label">{{ detail_field.label }}</label></td> <td><span>{{ detail_field.value }} {{ (detail_field.suffix) ? '%' }}</span></td>
-  </tr>
-  {% endif %}
-{% endfor %}
-</table>
+<h2 class="rate_plan__detail-card__title">Rate Card</h2>
```
jacine: There should be a wrapper for this `<div class="rate-plan-detail">` and then each field wrapper or element wrapper should be something like `rate-plan-detail__title`.

templates/rate-plan-detail.html.twig
```diff
@@ -24,44 +29,61 @@
...
+<div class="field field--inline">
```
jacine: Should also have `rate-plan-detail__metering-type` or `rate-plan-detail__type`.

templates/rate-plan-detail.html.twig
```diff
@@ -24,44 +29,61 @@
...
+  <div class="field__label">{{ "Rate card is based on"|t }}</div>
+  <div>{{ detail.meteringType }}</div>
+</div>
+<div class="field field--inline">
```
jacine: Should have `rate-plan-detail__duration`.

templates/rate-plan-detail.html.twig
```diff
@@ -24,44 +29,61 @@
...
-  <tr class="field--label-inline">
-    <td><label class="field__label">{{ detail_field.label }}</label></td> <td><span>{{ detail_field.value }} {{ (detail_field.suffix) ? '%' }}</span></td>
-  </tr>
-  {% endif %}
-{% endfor %}
-</table>
+<h2 class="rate_plan__detail-card__title">Rate Card</h2>
+<div class="field field--inline">
+  <div class="field__label">{{ "Rate card is based on"|t }}</div>
+  <div>{{ detail.meteringType }}</div>
```
jacine: Should have `<div class="field__item">`.

templates/rate-plan-detail.html.twig
```diff
@@ -24,44 +29,61 @@
...
+  <div class="field__label">{{ "Volume aggregation basis"|t }}</div>
+  <div>{{ detail.duration }} {{ detail.durationType|lower }}</div>
+</div>
+
+<div class="rate_plan__detail-card__overview">
+  {% for detail_field in detail_fields %}
+    {% if detail_field.value %}
+      <div class="field field--inline">
+        <div class="field__label">{{ detail_field.label }} </div>
+        <div>{{ detail_field.value }} {{ (detail_field.suffix) ? '%' }}</div>
```
jacine: Should have `<div class="field-item">`, and `<span class="field-suffix">{{ detail_field.suffix ? '%' }}</span>`.

templates/rate-plan-detail.html.twig
```diff
@@ -24,44 +29,61 @@
...
+    {% endif %}
+  {% endfor %}
+</div>
+
+{% if (ratecard_rates) %}
+  {% include 'ratecard-rates.html.twig' with {'rates': ratecard_rates} %}
+{% endif %}
+
+{% if (revshare_rates) %}
+  {% include 'revshare-rates.html.twig' with {'rates': revshare_rates} %}
```
jacine: See comment above.

templates/ratecard-rates.html.twig
```diff
@@ -0,0 +1,27 @@
...
+ * - detail:      See: "rate-plan-detail.html.twig"
+ * - rates:
+ *   - id:                          The unique ID of the rate plan rate.
+ *   - startUnit:                   The number the rate starts at.
+ *   - endUnit:                     The rate is good until this nuber of units.
+ *   - rate:                        The rate that is applies within this range.
+ */
+#}
+
+<div class="rate_plan__detail-card__ratecard-rates rate-list-wrapper">
```
jacine: Maybe something like `<div class="rate-plan-rates">`.  And really this template and the one below should just either be in the base hook template, or they should have their own theme hook.  Using `rate_plan_detail` as a base hook for this is odd. 

Then you can add another class, a modifier, so signify the type, e.g. `rate-plan-rates--ratecard` and in the template below `rate-plan-rates--revshare`.

templates/ratecard-rates.html.twig
```diff
@@ -0,0 +1,27 @@
...
+    <div class="row ratecard-rate__header row">
+        <div class="column"><strong>{{ "Volume bands"|t }}</strong></div>
+        <div class="column"><strong>{{ "Cost per API call"|t }}</strong></div>
+    </div>
+    {% for rate in rates %}
+        <div class="row ratecard-rate">
+            <div class="column">{{ "Greater than"|t }} {{ rate.startUnit }} {% if rate.endUnit %}{{ "up to"|t }} {{ rate.endUnit }}{% endif %}</div>
+            <div class="column">{{ rate.rate }}</div>
```
jacine: We definitely do not want to use `.row` and `.column`.   I still think these are best displayed in a table, since they are tabular data, but if you really want to add row/column CSS, these need to be prefixed.  `rate-plan-rates__row` and `rate-plan-rates__column`.

jacine (dismissed):
If you need help with this let me know.

templates/rate-plan-detail.html.twig
```diff
@@ -6,7 +6,12 @@
  * Variables:
  * - detail
  *   - id:
```
jacine: `id:` line has no docs.  Is `detail` the rate plan entity?   If so, maybe we should just call it `rate_plan`?

templates/rate-plan-detail.html.twig
```diff
@@ -6,7 +6,12 @@
...
+ *   - ratePlanRates:               All rate plan rate details, such as the type of rate plan (REVSHARE or RATECARD), the rate for a rate card plan, the revenue share for a revenue share plan, and the range (starting unit and ending unit for which the rate plan rate applies).
+ *     - id:                          The unique ID of the rate plan rate.
+ *     - startUnit:                   The number the rate starts at.
+ *     - endUnit:                     The rate is good until this nuber of units.
+ *     - rate:                        The rate that is applies within this range.
+ *     - revshare:                    The revenue share percentage.
  *   - aggregateFreemiumCounters:   Flag that specifies whether or not aggregate counters are enabled to determine if usage of an API product is in the free range. Aggregate counters must be enabled to set up a freemium plan for a product.
  *   - aggregateStandardCounters:   Flag that specifies whether or not aggregate counters are used to determine the band of usage (such as a volume band for a rate card plan).
  *   - aggregateTransactions:
```
jacine: There are no docs for this line either?  Maybe it's not needed?

templates/rate-plan-detail.html.twig
jacine: Should be translatable: `{{ 'Rate Card'|t }}`

apigee_m10n.libraries.yml
Jaesin: They are for two different fields which may or may not be configured to be displayed.

apigee_m10n.module
Jaesin: I agree, I'll left these in until I could confirm they won't be needed. I think the only thing that needs to be customized is "Rate card" vs "Rate card and Revenue" so that can just be done in the template.

apigee_m10n.module
Jaesin: `futurePlanLinks` is a link field. It isn't using a custom formatter so we don't have access tot he render array. Do you think it's worth creating a custom formatter in order to be able to avoid the pre-process?

jacine: No, I would put the CSS all in one library. In fact, I would put it the same file as the entity styles, because it’s part of the entity view, even if one of those fields might not be there. That makes only one library to override when these styles are not wanted instead of 3. If someone is really bothered by the tiny amount of CSS that’s in the aggregated version of this, they can override it, but I think it actually is less harm to keep it all in one in this case.  The reason being that some rare plan pages will have it and some won’t, so the aggregated bundle will be different depending on this field, which probably has more of an impact. 
 
But, TBH, I wouldn’t include any of this CSS in the module for these fields at all.  I would use a table for the tabular data (volume bands), and rely on field classes for the rest.

templates/rate-plan-detail.html.twig
Jaesin: OK, I moved them inline.

apigee_m10n.module
Jaesin: They have been removed.

templates/rate-plan-detail.html.twig
Jaesin: This has been fixed.

templates/rate-plan-detail.html.twig
Jaesin: This has been fixed.

templates/rate-plan-detail.html.twig
Jaesin: Fixed.

templates/rate-plan-detail.html.twig
Jaesin: I added `field__item` to all of the values.

templates/rate-plan-detail.html.twig
Jaesin: This wasn't really needed. I refactored this table without the loop.

templates/rate-plan-detail.html.twig
Jaesin: I removed the separate templates for the rate plan lists.

templates/ratecard-rates.html.twig
Jaesin: Cool. done.

templates/ratecard-rates.html.twig
Jaesin: I changed these to `rate-plan-rates__row` and `rate-plan-rates__column`.

templates/rate-plan-detail.html.twig
Jaesin: I cleaned up the docs for this file.

templates/rate-plan-detail.html.twig
Jaesin: Yeah, it's not being used by monetization so I removed it.

src/Plugin/Field/FieldFormatter/PriceFormatter.php
```diff
@@ -0,0 +1,190 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public function viewElements(FieldItemListInterface $items, $langcode) {
+    $options = $this->getFormattingOptions();
+    $elements = [];
+    foreach ($items as $delta => $item) {
+      $elements[$delta] = [
+        '#markup' => $this->currencyFormatter->format($item->amount, $item->currency_code, $options),
```
jacine, line 160: Should there be a space between the code and the amount? (Note: I changed the first one first one to show currency code)

Screen Shot 2019-06-25 at 12 41 30 PM

templates/rate-plan-detail.html.twig
```diff
@@ -24,44 +28,108 @@
...
+
+    <div class="revshare-rate rate-plan-rates__row">
+      <div class="rate-plan-rates__column">
+        <div class="field field--inline rate-plan-detail__overview__country">
+          <div class="field__label">{{ "Country"|t }}</div>
+          <div class="field__item">{{ detail.organization.country }}</div>
+        </div>
+      </div>
+      <div class="rate-plan-rates__column">
+        <div class="field field--inline rate-plan-detail__overview__pricing-type}">
```
jacine:
```diff
-<div class="field field--inline rate-plan-detail__overview__pricing-type}">
+<div class="field field--inline rate-plan-detail__overview__pricing-type">
```

templates/rate-plan-detail.html.twig
```diff
@@ -24,44 +28,108 @@
...
-  {% if detail_field.value %}
-  <tr class="field--label-inline">
-    <td><label class="field__label">{{ detail_field.label }}</label></td> <td><span>{{ detail_field.value }} {{ (detail_field.suffix) ? '%' }}</span></td>
-  </tr>
+  {% if entity.freemiumUnit or entity.freemiumDuration %}
+    <div class="field field--inline rate-plan-detail__freemium">
+      <div class="field__label">{{ "Free quantity"|t }}</div>
+      <div class="field__item">
+        {% if entity.freemiumUnit and entity.freemiumDuration and entity.freemiumDurationType %}
+          {{ "Up to"|t }} {{ entity.freemiumUnit }} {{ (entity.freemiumUnit == 1 ? "transaction for" : "transactions for")|t }} {{ entity.freemiumDuration }} {{ (entity.freemiumDuration == 1 ? entity.freemiumDurationType : entity.freemiumDurationType ~ 's')|lower|t }}.
```
jacine: These strings are really, really messy and difficult to grok.  Typically you are supposed to use `{% trans %}` with `{% plural %}` but having to run filters on this stuff too, make it even more complex than it already is, because you [can't even run filters](https://www.drupal.org/project/drupal/issues/2610436) inside trans. It would help if you at least did the `|lower` transform in the module code, and honestly the strings themselves, in the module code.

templates/rate-plan-detail.html.twig
Jaesin: I fixed this one.

templates/rate-plan-detail.html.twig
Jaesin: I moved this logic to the formatter. It's probably better since the real values here come from the rate plan entity and not the details property.

src/Plugin/Field/FieldFormatter/PriceFormatter.php
Jaesin: There are tree ways to configure the formatter and this is what they look like: 

image

If there are some improvements to be made, maybe we can address them in follow ups.

apigee_m10n.module
Jaesin: I think it makes sense to keep these separate and let aggregation handle it.

cnovak: @jacine seems to have a point about using one CSS file, it may be easier to theme and can be cached by browser if it is not changing on each page.

css/apigee_m10n.future-links-field.css
```diff
@@ -0,0 +1,21 @@
+.field--name-futureplanlinks {
```
cnovak, line 1: Like @jacine pointed out earlier, is it worth breaking out this small amount of CSS? Seems like it will be more costly than putting it all in one file.

src/Entity/RatePlan.php
```diff
@@ -228,6 +259,20 @@ public function toUrl($rel = 'canonical', array $options = []) {
     return $url;
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  protected function getFieldValue(string $field_name) {
+    // Add Use the price value to the field name for price items.
```
cnovak:
```suggestion
    // Add the price value to the field name for price items.
```

src/Entity/RatePlan.php
```diff
@@ -253,6 +298,113 @@ private function getUser() {
...
+    if (!isset($this->currentRatePlan)) {
+      // Check whether or not this plan has a parent plan.
+      if (($decorated = $this->decorated()) && $decorated instanceof RatePlanRevisionInterface) {
+        // Create a date to compare whether a plan is current. Current means the
+        // plan has already started and is a parent revision of this plan.
+        $today = new \DateTimeImmutable('today', $this->getStartDate()->getTimezone());
+        // The previous revision is our starting point.
+        $parent_plan = $decorated->getPreviousRatePlanRevision();
+        // Loop through parents until the current plan is found.
+        while ($parent_plan && ($today < $parent_plan->getStartDate() || $today > $parent_plan->getEndDate())) {
```
cnovak, line 391: I don't think you can ever have a situation where the previous rate plan is not a current plan, but good to have this logic.

src/Entity/Property/SetUpFeePropertyAwareDecoratorTrait.php
```diff
@@ -0,0 +1,54 @@
...
+ * with this program; if not, write to the Free Software Foundation, Inc., 51
+ * Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ */
+
+namespace Drupal\apigee_m10n\Entity\Property;
+
+/**
+ * Trait SetUpFeePropertyAwareDecoratorTrait.
+ */
+trait SetUpFeePropertyAwareDecoratorTrait {
```
cnovak, line 25: Are we planning on having multiple classes that need to have a setup fee? Seems like a lot of traits being set up versus leaving the these getters and setters in the RatePlan as you had it before.

cnovak:
@Jaesin & @arshad have signed CLA

src/Entity/Storage/RatePlanStorage.php
```diff
@@ -132,6 +145,35 @@ public function loadById(string $package_id, string $id): RatePlanInterface {
     return $entity;
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function loadFutureRatePlan(RatePlanInterface $ratePlan): ?RatePlanInterface {
+    if (!isset($this->future_rate_plans[$ratePlan->id()])) {
+      // Future plans haven't started as of today.
+      $today = new \DateTimeImmutable('today', $ratePlan->getStartDate()->getTimezone());
```
cnovak, line 154: `DateTimeImmutable()` can throw an exception.  Should we wrap, log an error, and return NULL if this happens?

src/Entity/Storage/RatePlanStorage.php
```diff
@@ -132,6 +145,35 @@ public function loadById(string $package_id, string $id): RatePlanInterface {
...
+      // Load all plans for this package.
+      $all_plans = $this->loadPackageRatePlans($ratePlan->getPackage()->id(), TRUE);
+      // Loop through to see if any future previous revisions are the given
+      // plan.
+      foreach ($all_plans as $future_candidate) {
```
cnovak, line 159: This is expensive, but I cannot think of a better way to do this since there doesn't seem to be an attribute to give you future rate plans for a given plan.

src/Plugin/Field/FieldType/PurchaseFieldItem.php
```diff
@@ -31,6 +31,7 @@
  *   id = "apigee_purchase",
  *   label = @Translation("Purchase rate plan"),
  *   description = @Translation("Purchase rate plan computed item."),
+ *   category = @Translation("Apigee"),
```
cnovak, line 34: "Apigee" isn't really a translatable string, but maybe you need to pass it in as translatable.

templates/rate-plan-detail.html.twig
```diff
@@ -4,12 +4,16 @@
...
  * Variables:
- * - detail
- *   - id:
- *   - ratePlanRates:               Rate plan rate details, such as the type of rate plan (REVSHARE or RATECARD), the rate for a rate card plan, the revenue share for a revenue share plan, and the range (starting unit and ending unit for which the rate plan rate applies).
+ * - detail                       The rate plan detail object.
+ *   - id:                          The unique ID for the rate plan detail.
+ *   - ratePlanRates:               All rate plan rate details, such as the type of rate plan (REVSHARE or RATECARD), the rate for a rate card plan, the revenue share for a revenue share plan, and the range (starting unit and ending unit for which the rate plan rate applies).
+ *     - id:                          The unique ID of the rate plan rate.
+ *     - startUnit:                   The number the rate starts at.
+ *     - endUnit:                     The rate is good until this nuber of units.
```
cnovak:
```suggestion
 *     - endUnit:                     The rate is good until this number of units.
```

templates/rate-plan-detail.html.twig
```diff
@@ -24,44 +28,100 @@
  *   - ratingParameterUnit:         Unit that applies to the ratingParameter.
  *   - revenueType:                 Basis of the revenue share in a revenue share plan. Valid values include: GROSS (based on percentage of the gross price of a transaction) and NET (based on a percentage of the net price of a transaction).
  *   - type:                        Revenue model for the plan. Valid values include: REVSHARE, RATECARD, REVSHARE_RATECARD, USAGE_TARGET (adjustable notifications).
+ * - ratecard_rates:             The RATECARD rates.
+ *   - id:                          The unique ID of the rate plan rate.
+ *   - startUnit:                   The number the rate starts at.
+ *   - endUnit:                     The rate is good until this nuber of units.
```
cnovak:
```suggestion
 *   - endUnit:                     The rate is good until this number of units.
```

cnovak (changes requested):
Overall questions and a few spelling fixes

src/Entity/Storage/RatePlanStorage.php
Jaesin: I agree it's not optimal. I would like to get the list for the package cached soon. Maybe we can have a followup for that.

src/Plugin/Field/FieldFormatter/RatePlanDetailsFormatter.php
```diff
@@ -71,13 +74,67 @@ public function viewElements(FieldItemListInterface $items, $langcode) {
...
+    // Get only rate card entries.
+    $ratecard_rates = array_filter($all_rates, function ($rate) {
+      return ($rate instanceof RatePlanRateRateCard);
+    });
+    // Get only revshare entries.
+    $revshare_rates = array_filter($all_rates, function ($rate) {
+      return ($rate instanceof RatePlanRateRevShare);
+    });
+
+    // The logic for the free quanity is a little complicated for the template.
```
cnovak:
```suggestion
    // The logic for the free quantity is a little complicated for the template.
```

cnovak (changes requested):
One more spelling fix

css/apigee_m10n.future-links-field.css
Jaesin: The reason they are separate is because the CSS in each is specific to a field which may or may not be displayed. CSS aggregation in Drupal 8 handles this kind of thing really well.

cnovak: Sounds good

src/Entity/RatePlan.php
Jaesin: I'de rather remove the logic to keep it simple if that is the case.

src/Entity/Property/SetUpFeePropertyAwareDecoratorTrait.php
Jaesin: I don't think there are any other entities using this property. We can change it back if you like.

src/Entity/Storage/RatePlanStorage.php
Jaesin: Sure. If there was an issue with the time zone, you would probably get an exception here.

src/Plugin/Field/FieldType/PurchaseFieldItem.php
Jaesin: I'm pretty sure it isn't required.

cnovak (approved):
Other comments can be dealt with followup issues if we decide to fix them

PR #138, opened 2019-06-21
[#103] Show already purchased message in purchased form

### LINKED ISSUE #103: The "already subscribed" check should be in the subscription form instead of the purchase widget.
https://github.com/apigee/apigee-m10n-drupal/pull/39 adds an "Already Subscribed" notice and hides the subscribe form when using the `apigee_subscribe_form` field formatter when rendering a rate plan. This logic should be moved to the "purchase" form so the rate plan subscribe page ( `/user/{user}/monetization/package/{package}/plan/{rate_plan}/subscribe`) can also benefit from this improvement.

shadcn:
Fixes #103


Jaesin:
Can we remove the duplicate logic from `\Drupal\apigee_m10n\Plugin\Field\FieldFormatter\PurchasePlanFormFormatter::viewValue()`?

Jaesin (approved):
LGTM

PR #141, opened 2019-06-26
[Fixes #139] Hide field types that aren't usable to other entities.

### LINKED ISSUE #139: Apigee field types are unusable to other entity types.
The `apigee_m10n` module defines some field types so it can use custom formatters for rendering. Other entities cannot use these fields but there is nothing to stop a user from adding them to a content entity.

**To Reproduce**

1. Go to 'admin/structure/types/manage/article/fields/add-field'
2. Select "Apigee Organization"
3. Click 'Save'

**Expected behavior**

Most of the fields aren't really useful to other entity types. In order to be useful they would need to be able to serialize and unserialize complex objects for storage and define the schema. It probably better to restrict their use to their intended purpose and not allow them to be added to other entity types.

Jaesin:


shadcn (approved):
This is good to go 👍

PR #146, opened 2019-06-27
[Fixes #65] Require the bcmath php extensions

### LINKED ISSUE #65: WSOD after enabling the module
It seems `commerceguys/intl` does not require bcmath extension in its composer.json because it is only required by [some classes that it contains](https://github.com/commerceguys/intl/search?q=bcmath&unscoped_q=bcmath). Although if this extension is not installed you get a WSOD after you enabled this module with this error in the logs:

```
  66   11/Mar 10:53   php      Error      RuntimeException: The bcmath extension is required by CurrencyFormatter. in CommerceGuys\Intl\Formatter\CurrencyFormatter->__construct() (line 81 of
 vendor/commerceguys/intl/src/Formatter/CurrencyFormatter.php).
```

Suggested solution:
As it seems this module depends on at least one class from `commerceguys/intl` that requires bcmath, add bcmath extension to this module's composer.json.

Jaesin:
since the formatters require `commerceguys/intl`.

### DIFF
--- a/apigee_m10n.install
+++ b/apigee_m10n.install
@@ -53,6 +53,16 @@ function apigee_m10n_requirements($phase) {
     }
   }
 
+  if ($phase == 'install' || $phase == 'runtime') {
+    if (!extension_loaded('bcmath')) {
+      $requirements['apigee_monetization_bcmath'] = [
+        'title' => t('BC Math'),
+        'description' => t('Apigee Monetization requires the BC Math PHP extension in order to format prices.'),
+        'severity' => REQUIREMENT_ERROR,
+      ];
+    }
+  }
+
   return $requirements;
 }
 
--- a/composer.json
+++ b/composer.json
@@ -5,6 +5,7 @@
     "description": "Apigee Edge Monetization for Drupal",
     "require": {
         "php": ">=7.1",
+        "ext-bcmath": "*",
         "cweagans/composer-patches": "~1.6",
         "commerceguys/intl": "~1.0",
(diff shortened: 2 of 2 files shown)


kscheirer (approved):
looks good to me

apigee_m10n.install
```diff
@@ -53,6 +53,16 @@ function apigee_m10n_requirements($phase) {
     }
   }
 
+  if ($phase == 'install' || $phase == 'runtime') {
```
cnovak, line 56: Tested new install, got: 
```
 Problem 1
    - Installation request for drupal/apigee_m10n dev-65_require-bcmath -> satisfiable by drupal/apigee_m10n[dev-65_require-bcmath].
    - drupal/apigee_m10n dev-65_require-bcmath requires ext-bcmath * -> the requested PHP extension bcmath is missing from your system.

```

Tested upgrading to this PR, get the same message.  I don't think we need this install/runtime check because you cannot install w/out having bcmath.  Can you think of a way around that I did not think of that would be a reason we add this code?

cnovak (changes requested):

apigee_m10n.install
Jaesin: If you require this module on your local that has bcmath and then deploy to a server that doesn't have bcmath, you would end up with an error and probably not know why.

cnovak: Good point, let's leave it in for that situation

cnovak (approved):

PR #147, opened 2019-06-28
Add default add credit permissions for authenticated users

arlina-espinoza:
When installing the _"apigee_m10n_add_credit"_ module, no permissions are set by default. I think it would make for a better experience to give the authenticated user the permission to _"add credit to own developer prepaid balance"_. Likewise, when installing _"apigee_m10n_teams"_, give the authenticated user the _"add credit to own team prepaid balance" permission"_.


jacine (approved):
Tested this and it works perfectly.  Thanks you @arlina-espinoza!

PR #152, opened 2019-07-02
[Fixes #149] Rename `package` entity to `product_bundle`

### LINKED ISSUE #149: Packages should be renamed Product Bundle
The `Package` entity is now called `Product Bundle` in the Edge user interface and documentation.  The API documentation has a link out to [a page talking about making API calls to view product bundles , but the API is still /packages](https://docs.apigee.com/api-platform/monetization/api-product-bundles#addprodtopackapi).

The entity should be renamed.  Comments should be added to the entity to explain that the API is still called packages to reduce confusion.

Jaesin:


apigee_m10n.links.task.yml
```diff
@@ -4,10 +4,10 @@ apigee_m10n.settings.rate_plan:
   base_route: apigee_m10n.settings.rate_plan
   weight: -10
 
-entity.package.collection:
-  title: 'List packages'
-  route_name: entity.package.collection
-  base_route: apigee_m10n.settings.package
+entity.product_bundle.collection:
+  title: 'List product bundles'
```
shadcn: Nit: Should we keep this to just `List` like core does?

Screenshot_7_10_19__12_59_PM

shadcn (approved):
Tested. Looks good to me. Approved.

apigee_m10n.links.task.yml
Jaesin: That's a good suggestion. Thanks.

Issue #150, opened 2019-06-28
"Add credit" button on prepaid balance page does not display


cnovak:
**Describe the bug**
The add credit button does not show up on the balance page until after you have added money to your account, but you cannot add money w/out having the button (catch-22 situation).

**To Reproduce**
1. Follow [Step 4: Add an “Add Credit” button to your prepaid balance page](https://www.drupal.org/docs/8/modules/apigee-monetization/enable-the-apigee-monetization-add-credit-module#step-4-add-an-add-credit-button)

**Expected behavior**
Button is displayed

**Actual behavior**
Button is not displayed. However, if you already have a prepaid balance, button will be displayed.

**Screenshots**
<img width="1005" alt="no-button" src="https://user-images.githubusercontent.com/613093/60368344-7ef2b680-99a5-11e9-867a-ce47a44f58e5.png">

**Version Info**
RC-1

jacine:
I think this was fixed by #155.  Feel free to reopen if that's wrong.

PR #155, opened 2019-07-03
[#150] Show "Add credit" button on balances even if no balances available

arlina-espinoza:
Added the "add credit" button for all currencies enabled. 
Example for when no balance available for any currency:
Screenshot_2019-07-03 Prepaid balance Apigee Developer Portal Kickstart M10N(2)

Example for when a balance is available for only some currencies:
Screenshot_2019-07-03 Prepaid balance Apigee Developer Portal Kickstart M10N(1)


jacine (approved):
Works great!  Thank you @arlina-espinoza 👏 

Here are a couple of screenshots.

## User with No credit
add-credit-user-with-no-credit

## User with USD credit
add-credit-user-with-USD-balance

Jaesin (approved):
Looks good. Tested with and without the add credit product being configured.

PR #159, opened 2019-07-08
[#151] Show all rate plan types including developer and developer category

### LINKED ISSUE #151: Developer/Developer categories are not displayed
**Describe the bug**
Plans can be created that are only displayed for a developer or developer category.  This module is not displaying those plans.

**To Reproduce**
1. In Edge, create a plan with the audience "Developer category" or "Developer"
2. Log in as that developer or a developer/company in that developer category
3. Look for the plan you created, notice it is not displayed

**Expected behavior**
The plan should be displayed

**Additional context**
The calls being made to the Edge API are:
`https://api.enterprise.apigee.com/v1/mint/organizations/{org}/monetization-packages/{package/productbundle}/rate-plans`, but this will not show developer or developer category plans.

The call needs to be `https://api.enterprise.apigee.com/v1/mint/organizations/{org}/monetization-packages/{package/productbundle}/developers/{email}/rate-plans`

Also note that to put a developer in a developer category, you need to make an API call. Do a **GET** call to https://api.enterprise.apigee.com/v1/organizations/:org_name/developers/:email, then do a **PUT** to the same resource adding the **MINT_DEVELOPER_CATEGORY** custom attribute.  The [docs about this say](https://docs.apigee.com/api-platform/monetization/manage-developer-categories#api) to to a **PUT** to https://api.enterprise.apigee.com/v1/**mint**/organizations/{org_name}, but you need to make a **PUT** to the non-mint resource.

**Version Info**
all

shadcn:
Fixes #151


Jaesin (commented):
Developer rate plans should only be accessible by the assigned developer so we would need some additional access control and some additional tests. One for a the developer the plan is for and another to test access is denied to other developers.

If a rate plan is of the type "RatePlanInterface::TYPE_DEVELOPER_CATEGORY", the developer must have the developer category ID in it's `MINT_DEVELOPER_CATEGORY` attribute.
So we should have a check for `DEVELOPER_CATEGORY` plans in the rate plan access control as well as the tests for that.

For a developer rate plan, the developer could be a Company. Access to a company rate plan would have to be handled by the teams module. If a developer is part of a team that has access via the teams developer category, team members should also have access to the rate plan. I will create a separate ticket for allowing access to a rate plan for team members.

src/Entity/Access/RatePlanAccessControlHandler.php
```diff
@@ -0,0 +1,107 @@
...
+      if ($developer = $rate_plan->getDeveloper()) {
+        return AccessResult::allowedIf($account->getEmail() === $developer->getEmail())
+          ->andIf(AccessResult::allowedIfHasPermission($account, "$operation rate_plan"));
+      }
+      return AccessResult::forbidden("User {$developer->getEmail()} cannot view developer rate plan.");
+    }
+
+    return $access->andIf(AccessResult::allowedIfHasPermission($account, "$operation rate_plan"));
+  }
+
```
Jaesin, line 106: This could something like: 

```
  /**
   * {@inheritdoc}
   */
  protected function checkCreateAccess(AccountInterface $account, array $context, $entity_bundle = NULL) {
    return AccessResult::forbidden('Rate plans cannot be created via the developer portal.');
  }
```

src/Entity/Access/RatePlanAccessControlHandler.php
```diff
@@ -0,0 +1,107 @@
+<?php
+
+/*
+ * Copyright 2018 Google Inc.
```
Jaesin: Should we be using 2019 here?

arlina-espinoza (approved):
Looks good :) Tested it with a developer plan, and a developer category plan. In both cases it works as expected displaying the appropriate rate plans after a cache rebuild.

src/Entity/Access/RatePlanAccessControlHandler.php
shadcn: Thanks. Added.

src/Entity/Access/RatePlanAccessControlHandler.php
shadcn: Updated. Thanks.

Issue #157, opened 2019-07-04
Use controllers from the SDK


Jaesin:
Billing report controllers were added to the SDK here: https://github.com/apigee/apigee-client-php/pull/51 but there are still a couple of SDK controllers in the code base. `\Drupal\apigee_m10n\SDK\Controller\BillingDocumentsController` and `\Drupal\apigee_m10n\SDK\Controller\PrepaidBalanceReportsController`.

This task is to replace and remove the controllers from the code base.

An alternative would be to mark these controllers as internal and deprecated for the 1.0 release and remove them in the future.

PR #164, opened 2019-07-10
[#157] Use the prepaid balance controller from the PHP SDK.

Jaesin:
This is only a partial fix for #157. The `\Drupal\apigee_m10n\SDK\Controller\BillingDocumentsController` controller still needs to be eliminated but will be handled in a separate PR.


shadcn (approved):
Reviewed and tested this manually. Looks good to me. Thanks @Jaesin. Approved.

PR #165, opened 2019-07-11
[Fixes #157] Remove the billing documents controller in favor of using past x months.

### LINKED ISSUE #157: Use controllers from the SDK

Jaesin:
Since the `billing-documents-months` endpoint only shows open billing months it isn't that useful for prepaid billing statements. If the endpoint was developer specific and returned only month that there was transactions for the developer it would be more useful but that's not what it does. It only returns billing months that are still of open status for all developers.

Since there isn't an easy way to determine which months contain transaction data,wWe decided to add a setting to show the past x number of months. We add an additional check to make sure we aren't allowing months from before the org was created since there won't be any data from before then.


arlina-espinoza (approved):
Reviewed and tested, and works as expected.
The only thing I added is a form rebuild to the prepaid statements form to keep the last selected options, as it gets annoying to have to re-select everything in case a particular year/month doesn't have a report.

PR #170, opened 2019-07-17
Fix the add credit config path.

Jaesin:


shadcn (commented):
Hmm weird. I'm seeing a 404 on this path. Not sure why this is not working for me. The tabs and links are gone as well.

arlina-espinoza (approved):
Looks good! Thanks @Jaesin !

Issue #162, opened 2019-07-09
Allow access to team specific rate plan and team category rate plans.


Jaesin:
A developer should have access to a `DEVELOPER` rate plan if the developer is a member of the team the rate plan was created for. A developer should also have access to a `DEVELOPER_CATEGORY` rate plan if the developer is a part of a team that has the rate plans developer category id set in the teams `MINT_DEVELOPER_CATEGORY` attribute.

**Describe the solution you'd like**
There should be a rate plan entity access check for view and subscribe operations that checks for team access for developer and developer category rate plans. Their is already a "View rate plans" and a "Purchase a rate plan" team permission.

There should be at lease 8 new access scenarios in the tests.

- A developer can view a `DEVELOPER` rate plan because of team membership.
- A developer can not view a `DEVELOPER` rate plan because of a lack of team membership.
- A developer can subscribe to a `DEVELOPER` rate plan because of team membership.
- A developer can not subscribe to a `DEVELOPER` rate plan because of a lack of team membership.
- A developer can view a `DEVELOPER_CATEGORY` rate plan because of team membership.
- A developer can not view a `DEVELOPER_CATEGORY` rate plan because of a lack of team membership.
- A developer can subscribe to a `DEVELOPER_CATEGORY` rate plan because of team membership.
- A developer can not subscribe to a `DEVELOPER_CATEGORY` rate plan because of a lack of team membership.

arlina-espinoza:
Fixed in #179 .

PR #179, opened 2019-07-20
[#162] Allow access to team specific rate plan and team category rate plans

arlina-espinoza:
Regarding [#162]: Adds access control to company rate plans, and category rate plans where the "category" attribute is inherited from the company/team.


shadcn (approved):
This is good to go. Thanks Arlina.

PR #182, opened 2019-07-24
[#175] Admin access for rate plans

### LINKED ISSUE #175: Admins can subscribe to a developer rate plan that was not created for them.
When viewing rate plans as a developer (i.e. /user/1/plans) the developer should not be able to purchase a rate plan unless they are the assigned developer or have the developer category.

**To Reproduce**

1. Create a developer rate plan assigned to a developer. 
2. View the rate plan as an admin that is not the assigned developer.
3. Observe you are able to purchase the rate plan.

**Expected behavior**

An admin would not be able to purchase a rate plan that is assigned to a different developer. It might be best if the plan developer context was required to view the rate plan to avoid confusion. If an admin can view the rate plan with the assigned developers context (via `/user/{assigned_developer_uid}/plans`), that should be enough.

Relates to #151.

shadcn:
Fixes #175


src/Entity/Access/RatePlanAccessControlHandler.php
```diff
@@ -72,14 +73,14 @@ protected function checkAccess(EntityInterface $entity, $operation, AccountInter
     /** @var \Drupal\apigee_m10n\Entity\RatePlanInterface $entity */
     $access = parent::checkAccess($entity, $operation, $account);
 
+    /** @var \Apigee\Edge\Api\Monetization\Entity\RatePlanInterface $rate_plan */
+    $rate_plan = $entity->decorated();
+
     // Allow access if user has permission `view/purchase rate_plan as anyone`.
-    if ($account->hasPermission("$operation rate_plan as anyone")) {
+    if ($rate_plan instanceof StandardRatePlanInterface && $account->hasPermission("$operation rate_plan as anyone")) {
```
arlina-espinoza, line 80: I don't think this is correct, as it would only allow admins to do operations on behalf of someone else on standard rate plans.

arlina-espinoza (changes requested):
While this PR will effectively block the purchase of a purchase the rate plan to an admin that is not the assigned developer, it will also give and access denied when an admin is trying to view/purchase a plan for someone else. Example:
- Rate plan "AC Plan 2" is assigned to a developer "dev-user", with drupal uid 3.
- As a drupal admin (uid: 2):
  - Tries to purchase the plan for itself (/user/2/monetization/product-bundle/{product_bundle}/plan/{rate_plan}: gets access denied - _OK_ because uid 2 is not the assigned developer and should not be able to purchase the plan.
  - Tries to purchase the plan for the assigned developer (/user/3/monetization/product-bundle/{product_bundle}/plan/{rate_plan}: PR returns access denied, but it should be allowed as uid 3 is the assigned developer.

It looks like it's not so much of an issue of access control - an admin with the `"$operation rate_plan as anyone"` permission should be able to access all of the plans-. Perhaps we can approach this by filtering out the plans that can effectively  be purchased by each user on it's `/user/{user}/plans` page?

src/Entity/Access/RatePlanAccessControlHandler.php
shadcn: @arlina-espinoza is this taken care of in https://github.com/apigee/apigee-m10n-drupal/pull/185?

arlina-espinoza: @arshad Yes, it would be#185 instead of this PR.

shadcn:
@arlina-espinoza I believe this can be closed now?

PR #184, opened 2019-07-25
Fix hidden errors in tests.

Jaesin:
There were a couple of errors that were hidden.

This also removes a redundant twig template.


arlina-espinoza (approved):
Looks good, thanks @Jaesin! Just curious to know which _"errors that were hidden"_?

PR #185, opened 2019-07-31
#175 Fix admins can subscribe to a developer rate plan that was not created for them

### LINKED ISSUE #175: Admins can subscribe to a developer rate plan that was not created for them.

arlina-espinoza:
Fixes #175. On a plans and pricing page it will filter out rate plans (user/{user}/plans) that the user cannot subscribe to (using the account from the URL, not the logged-in user, as that might be different). An analogous check is used on the teams plans page as well (/teams/{team}/plans).


shadcn (approved):
Actually I tested this a few days back but forgot to add the review here. This looks good. LGTM.

Issue #178, opened 2019-07-19
Review the add credit sub module


arlina-espinoza:
[Follow up of #169] Do a through review of the add credit sub module. Some things we may want to review:

- [Administrative usability] Admin /admin/config/apigee-edge/monetization/add-credit.
- [Administrative usability] Configuring add credit products.
- [Usability] The way the add credit product is displayed could use some review for possible improvements.
- [Implementation] When the purchased credit is added to the account it would be nice if we can report back to the order that the balance was successfully added or at least have an admin view of balance adjustments. The BalanceAdjustmentJob might need to be abandoned but that's probably OK since the job can't be retried anyway.

Some improvements could break backwards compatibility such as changes in how add credit products are configured.

arlina-espinoza:
**Administrative usability**
- The main pain points of _administrative usability_ have to do with all of the required configuration that needs to be done before actually configuring and using this module (set up a store, configure currencies supported in Edge, creating the products). The Kickstart install profile does a nice job of setting things up transparently, and it would be nice to move/replicate that behavior through a module setup form - see https://github.com/apigee/apigee-devportal-kickstart-drupal/issues/196#issuecomment-515497393. 
- A report of "add credit" transactions reporting if the balance adjustment was successfully added would be a nice added feature, instead of relying solely on the email notifications in case of errors. This would mean keeping track of commerce orders and Apigee transactions IDs.

**General usability**
- The field label "Unit price" is not very clear. Kickstart renames this field to "Amount to be added to your account balance", and that seems better. Validation messages and tests will need to be updated to match as needed. 
- The "add credit" button takes a long time as a user with  _"'add credit to any developer prepaid balance'"_ to show the modal - will need to see if someone else is running into this too.

**Others**
- I doubt the "add credit" product configuration will change, so I don't think there will be breaking changes moving forward.

arlina-espinoza:
Created issues for each item, closing.

PR #186, opened 2019-07-31
[#187] Update the "unit price" label in the add credit submodule.

arlina-espinoza:
Follow up on #178 : 

> The field label "Unit price" is not very clear. Kickstart renames this field to "Amount to be added to your account balance", and that seems better. Validation messages and tests will need to be updated to match as needed.


arlina-espinoza:
Closes #187 .

shadcn (approved):
I ran this locally and everything looks good. We can merge this when it is green again. Thanks @arlina-espinoza 

We might need a follow-up issue on kickstart to remove the deprecated code.

shadcn:
> We might need a follow-up issue on kickstart to remove the deprecated code.

I created https://github.com/apigee/apigee-devportal-kickstart-drupal/issues/244

PR #189, opened 2019-07-31
[#188] Add an "add credit" transaction log report.

### LINKED ISSUE #188: Admin report of "add credit" transactions
Follow up on #178 :

> A report of "add credit" transactions reporting if the balance adjustment was successfully added would be a nice added feature, instead of relying solely on the email notifications in case of errors. This would mean keeping track of commerce orders and Apigee transactions IDs.

arlina-espinoza:
Closes #188 :

> A report of "add credit" transactions reporting if the balance adjustment was successfully added would be a nice added feature, instead of relying solely on the email notifications in case of errors. This would mean keeping track of commerce orders and Apigee transactions IDs.

This PR adds a transaction log report of Apigee Transaction IDs and commerce order IDs. Because the API call when adding a prepaid balance does not return the transaction ID, it attempts to recover it by means of the last transaction recorded for the given developer/team in the prepaid balance report (the same report used for generating the "previous prepaid statements on _user/{user}/monetization/billing_).

Report is available under _"Reports > Add Credit Transactions Log"_ (/admin/reports/add-credit-transactions) to users with the permission _"'View Add Credit Log"_.

image


shadcn (approved):
Approved. Thanks @arlina-espinoza

PR #191, opened 2019-08-02
[#190] Adding configuration requirements for admin usability.

### LINKED ISSUE #190: Add requirement module for easier setup after install
Follow up of #178:
`The main pain points of administrative usability have to do with all of the required configuration that needs to be done before actually configuring and using this module (set up a store, configure currencies supported in Edge, creating the products). The Kickstart install profile does a nice job of setting things up transparently, and it would be nice to move/replicate that behavior through a module setup form - see apigee/apigee-devportal-kickstart-drupal#196 (comment).`

arlina-espinoza:
Closes #190 .
In summary, it uses the _requirement_ module to list the "add credit" module configuration requirements as a nice report with links and configuration forms.
Note: This brings over part of the work that was done for https://github.com/apigee/apigee-devportal-kickstart-drupal/pull/241 into the `apigee_m10n_add_credit` module.
image

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,7 +9,7 @@
         "cweagans/composer-patches": "~1.6",
         "commerceguys/intl": "~1.0",
         "drupal/apigee_edge": "~1.0",
-        "drupal/requirement": "1.x-dev",
+        "drupal/requirement": "~1.0",
         "apigee/apigee-client-php": "~2.0.3"
     },
     "require-dev": {
--- a/composer.json
+++ b/composer.json
@@ -9,6 +9,7 @@
         "cweagans/composer-patches": "~1.6",
         "commerceguys/intl": "~1.0",
         "drupal/apigee_edge": "~1.0",
+        "drupal/requirement": "1.x-dev",
         "apigee/apigee-client-php": "~2.0.3"
     },
     "require-dev": {
--- a/modules/apigee_m10n_add_credit/src/ApigeeEdgeSdkConnectorTrait.php
+++ b/modules/apigee_m10n_add_credit/src/ApigeeEdgeSdkConnectorTrait.php
@@ -1,0 +1,31 @@
+<?php
+
+namespace Drupal\apigee_m10n_add_credit;
+
+/**
+ * Provides a trait with utility method to fetch Apigee's Edge sdk connector.
+ */
+trait ApigeeEdgeSdkConnectorTrait {
+
+  /**
+   * The Apigee Edge SDK connector.
+   *
+   * @var \Drupal\apigee_edge\SDKConnectorInterface
+   */
+  protected $apigeeEdgeSdkConnector;
+
+  /**
+   * Gets the Apigee Edge SDK connector.
+   *
+   * @return \Drupal\apigee_edge\SDKConnectorInterface|mixed
+   *   The Apigee Edge SDK connector.
+   */
+  public function getApigeeEdgeSdkConnector() {
+    if (!$this->apigeeEdgeSdkConnector) {
+      $this->apigeeEdgeSdkConnector = \Drupal::service('apigee_edge.sdk_connector');
+    }
+
+    return $this->apigeeEdgeSdkConnector;
+  }
+
+}
--- a/modules/apigee_m10n_add_credit/src/Plugin/Requirement/Requirement/AddCreditProducts.php
+++ b/modules/apigee_m10n_add_credit/src/Plugin/Requirement/Requirement/AddCreditProducts.php
@@ -22,7 +22,7 @@ namespace Drupal\apigee_m10n_add_credit\Plugin\Requirement\Requirement;
 use Apigee\Edge\Api\Monetization\Controller\SupportedCurrencyController;
 use Apigee\Edge\Api\Monetization\Entity\SupportedCurrencyInterface;
 use CommerceGuys\Intl\Currency\CurrencyRepository;
-use Drupal\apigee_m10n_add_credit\ApigeeEdgeSdkConnectorTrait;
+use Drupal\apigee_m10n\ApigeeEdgeSdkConnectorTrait;
 use Drupal\apigee_m10n_add_credit\AddCreditConfig;
 use Drupal\commerce_price\Price;
 use Drupal\Core\Form\FormStateInterface;
--- a/modules/apigee_m10n_add_credit/src/Plugin/Requirement/Requirement/AddCreditProducts.php
+++ b/modules/apigee_m10n_add_credit/src/Plugin/Requirement/Requirement/AddCreditProducts.php
@@ -88,7 +88,7 @@ class AddCreditProducts extends RequirementBase implements ContainerFactoryPlugi
   protected $importableCurrencies;
 
   /**
-   * ApigeeEdgeConnection constructor.
+   * AddCreditProducts constructor.
    *
    * @param array $configuration
    *   A configuration array containing information about the plugin instance.
--- a/modules/apigee_m10n_add_credit/src/Plugin/Requirement/Requirement/AddCreditProducts.php
+++ b/modules/apigee_m10n_add_credit/src/Plugin/Requirement/Requirement/AddCreditProducts.php
@@ -42,6 +42,7 @@ use Symfony\Component\DependencyInjection\ContainerInterface;
  *   dependencies={
  *      "apigee_edge_connection",
  *      "commerce_store",
+ *      "add_credit_product_type",
  *   }
  * )
  */
--- a/modules/apigee_m10n_add_credit/src/Plugin/Requirement/Requirement/AddCreditProducts.php
+++ b/modules/apigee_m10n_add_credit/src/Plugin/Requirement/Requirement/AddCreditProducts.php
@@ -32,6 +32,8 @@ use Drupal\requirement\Plugin\RequirementBase;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 
 /**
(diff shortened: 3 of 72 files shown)


shadcn (approved):
This looks good to me. Thanks @arlina-espinoza

arlina-espinoza:
Thanks @arshad. Just waiting on feedback from @cnovak regarding adding the _requirement_ module dependency on both this PR and [Kickstart #233](https://github.com/apigee/apigee-devportal-kickstart-drupal/pull/241).

arunz6161:
Blocked by repo creation.

arlina-espinoza:
To do: Add the drupal/requirement module to this module.

modules/apigee_m10n_add_credit/src/ApigeeEdgeSdkConnectorTrait.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+namespace Drupal\apigee_m10n_add_credit;
+
+/**
+ * Provides a trait with utility method to fetch Apigee's Edge sdk connector.
+ */
+trait ApigeeEdgeSdkConnectorTrait {
```
shadcn: @arlina-espinoza Nice 👍

shadcn (approved):
@arlina-espinoza This all looks good to me. Tested locally on a kickstart and non-kickstart install. LGTM (when green)

shadcn:
@arlina-espinoza 

- I moved the Edge connection requirements to apigee_m10n since this is a requirement for apigee_m10n. 

- I also added a requirement for configuring the form display on order item. See Step 4 of https://www.drupal.org/docs/8/modules/apigee-monetization/enable-the-apigee-monetization-add-credit-module#step-3-add-products-to-the-store

Requirements_report___Apigee

composer.json
```diff
@@ -9,6 +9,7 @@
         "cweagans/composer-patches": "~1.6",
         "commerceguys/intl": "~1.0",
         "drupal/apigee_edge": "~1.0",
+        "drupal/requirement": "1.x-dev",
```
cnovak: We shouldn't be adding a dev version of a module. Should also move to 1.0 as soon as possible

cnovak (changes requested):
Need to use a stable version of requirement that comes from apigee/requirement-drupal project.

cnovak:
CLAs have been signed

modules/apigee_m10n_add_credit/src/Plugin/Requirement/Requirement/AddCreditProducts.php
```diff
@@ -0,0 +1,278 @@
...
+
+  /**
+   * An array of importable currencies.
+   *
+   * @var array|\CommerceGuys\Intl\Currency\Currency[]
+   */
+  protected $importableCurrencies;
+
+  /**
+   * ApigeeEdgeConnection constructor.
```
cnovak: Wrong class name

modules/apigee_m10n_add_credit/src/Plugin/Requirement/Requirement/CommerceStore.php
```diff
@@ -0,0 +1,158 @@
...
+
+    $form['address']['country_code'] = [
+      '#title' => $this->t('Country'),
+      '#type' => 'address_country',
+      '#required' => TRUE,
+      '#default_value' => $this->getConfigFactory()->get('system.date')->get('country.default'),
+    ];
+
+    $address_fields = [
+      AddressField::ADDRESS_LINE1 => [
```
cnovak, line 171: We should follow same logic as kickstart create store form where [we use org defaults](https://github.com/apigee/apigee-devportal-kickstart-drupal/blob/768e0540ce5f2d212023a2930571393f32961491/src/Installer/Form/ApigeeMonetizationConfigurationForm.php#L311). Would be good to not replicate this form/code  in two places, any ideas?

modules/apigee_m10n_add_credit/src/Plugin/Requirement/Requirement/CommerceStore.php
```diff
@@ -0,0 +1,158 @@
...
+    $form_state->cleanValues();
+
+    try {
+      $values = $form_state->getValues();
+      $store = $this->getEntityTypeManager()->getStorage('commerce_store')
+        ->create($values);
+      $store->save();
+    }
+    catch (\Exception $exception) {
+      watchdog_exception('apigee_kickstart', $exception);
```
cnovak: Wrong exception category

src/Plugin/Requirement/Requirement/ApigeeEdgeConnection.php
```diff
@@ -0,0 +1,66 @@
...
+ *   group="apigee_edge",
+ *   label="Apigee Edge connection",
+ *   description="A working connection to Apigee Edge is required.",
+ *   form="\Drupal\apigee_edge\Form\AuthenticationForm",
+ *   action_button_label="Configure credentials",
+ *   severity="error",
+ *   weight=-100
+ * )
+ */
+class ApigeeEdgeConnection extends RequirementBase {
```
cnovak, line 39: Shouldn't this requirement check be in the Apigee Edge module instead? Or this will need to be duplicated in that module, or am I missing something?

src/ApigeeEdgeSdkConnectorTrait.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+namespace Drupal\apigee_m10n;
+
+/**
+ * Provides a trait with utility method to fetch Apigee's Edge sdk connector.
+ */
+trait ApigeeEdgeSdkConnectorTrait {
```
cnovak, line 8: Shouldn't this be in Apigee Edge module? This class does not have anything monetization specific.

src/Plugin/Requirement/Requirement/ApigeeEdgeDeveloperSync.php
```diff
@@ -0,0 +1,68 @@
...
+ *   description="Sync Drupal users with Apigee Edge developers.",
+ *   severity="recommendation",
+ *   form="\Drupal\apigee_edge\Form\DeveloperSyncForm",
+ *   action_button_label="Sync developers",
+ *   dependencies={
+ *      "apigee_edge_connection",
+ *   }
+ * )
+ */
+class ApigeeEdgeDeveloperSync extends RequirementBase {
```
cnovak, line 41: Same with this class, this seems to be something that should be part of Apigee Edge module.

cnovak (changes requested):

src/Plugin/Requirement/Requirement/ApigeeEdgeConnection.php
shadcn: Yes. Agreed. But we have not decided yet if we want requirements to be a dependency for `apigee_edge` so for now this lives in `apigee_m10n` but yes ideally this would be in Edge.

src/ApigeeEdgeSdkConnectorTrait.php
shadcn: See https://github.com/apigee/apigee-m10n-drupal/pull/191#discussion_r371066752

src/Plugin/Requirement/Requirement/ApigeeEdgeDeveloperSync.php
shadcn: See https://github.com/apigee/apigee-m10n-drupal/pull/191#discussion_r371066752

modules/apigee_m10n_add_credit/src/Plugin/Requirement/Requirement/AddCreditProducts.php
shadcn: Nice catch. Thank you 👍

modules/apigee_m10n_add_credit/src/Plugin/Requirement/Requirement/CommerceStore.php
shadcn: Fixed.

modules/apigee_m10n_add_credit/src/Plugin/Requirement/Requirement/CommerceStore.php
shadcn: Looking into this. Trying to figure out if there's a way we could mark requirements as an installation step.

cnovak: We can break this out into a separate issue, but should make sure logic is the same in both places in this PR.

src/Plugin/Requirement/Requirement/ApigeeEdgeConnection.php
cnovak: Makes sense, we can move if needed.

modules/apigee_m10n_add_credit/src/Plugin/Requirement/Requirement/CommerceStore.php
shadcn: OK I'll make the changes to add orgs default and we can create a separate issue for the install tasks.

shadcn: This is done and I've create an issue to handle install tasks here: https://github.com/apigee/requirement-drupal/issues/3

cnovak (approved):
LGTM

PR #196, opened 2019-09-03
#116 Tests for team subscription

### LINKED ISSUE #116: Add tests for team subscription
PR #47  was created for this work without test coverage, need to write tests for this work.

arlina-espinoza:
Closes #116 . Tests the team purchased rate plan both as a member and non-member with different permissions, as well as the purchased rate plan listing view.


modules/apigee_m10n_teams/src/Entity/TeamsPurchasedPlan.php
```diff
@@ -113,25 +114,30 @@ public function getTermsAndConditions(): bool {
     }
   }
 
-  /**
-   * {@inheritdoc}
-   */
-  public function getTeam() {
```
Jaesin: Since this is public it should be deprecated instead of removed. It is also exposed as a field in `/config/apigee-monetization/purchased-plan-settings/display`. I think this is the only method that references `TeamsPurchasedPlan::getTeamReference()` so that code could just be folded in.

modules/apigee_m10n_teams/src/Entity/TeamsPurchasedPlan.php
```diff
@@ -169,6 +175,20 @@ public function purchasedPlanType() {
       : static::PURCHASED_PLAN_TYPE_DEVELOPER;
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function getOwner() {
```
Jaesin, line 191: Good catch.

modules/apigee_m10n_teams/src/MonetizationTeams.php
```diff
@@ -172,7 +172,7 @@ public function fieldWidgetInfoAlter(array &$info) {
    */
   public function purchasedPlanAccess(EntityInterface $entity, $operation, AccountInterface $account) {
     /** @var \Drupal\apigee_m10n_teams\Entity\TeamsPurchasedPlanInterface $entity */
-    if ($entity->isTeamPurchasedPlan() && ($team = $entity->get('team')->entity)) {
+    if ($entity->isTeamPurchasedPlan() && ($team = $entity->getTeamEntity())) {
```
Jaesin, line 175: I'm OK with adding `getTeamEntity()` but these should both do the same thing.

modules/apigee_m10n_teams/tests/response-templates/monetization-company.json.twig
```diff
@@ -0,0 +1,35 @@
...
+ *   - organization:  Flag that specifies whether the revenue is based on net.
+ *   - status:        Status indicator for the API product. Valid values include: CREATED, ACTIVE, INACTIVE. This value is maintained but not currently used by the API product.
+ *   - type:          Developer type (TRUSTED or UNTRUSTED).
+ */
+#}
+{
+    "billingType" : "PREPAID",
+    "broker" : false,
+    "customAttributes" : [ ],
+    "email" : "org###{{ developer.id }}",
```
Jaesin: What is `org###` for?

modules/apigee_m10n_teams/tests/response-templates/monetization-company.json.twig
```diff
@@ -0,0 +1,35 @@
+{#
+/**
+ * @file
+ *   Developer
+ *
+ * Usage:
+ *   @code {% include 'monetization-developer.json.twig' %} @endcode
+ *
+ * Variables:
+ * - developer: The user account for the developer.
```
Jaesin: I wonder why this wouldn't be renamed to company or team.

Jaesin (commented):
The tests all look good there is just a couple of things.

modules/apigee_m10n_teams/src/Entity/TeamsPurchasedPlan.php
arlina-espinoza: I added back this method so that it continues to be exposed as a field in the display settings page.

modules/apigee_m10n_teams/src/MonetizationTeams.php
arlina-espinoza: This should work the same.

modules/apigee_m10n_teams/tests/response-templates/monetization-company.json.twig
arlina-espinoza: Good catch, renamed to company.

modules/apigee_m10n_teams/tests/response-templates/monetization-company.json.twig
arlina-espinoza: The monetization API response for the company has a fake email in the form `{org_id}###{{ company.id }}@APIGEE-GEN.FOR-COMPANY` - because teams/companies do no have emails.

arunz6161:
Blocked on https://github.com/apigee/apigee-m10n-drupal/pull/202

modules/apigee_m10n_teams/tests/src/Traits/ApigeeMonetizationTeamsTestTrait.php
```diff
@@ -0,0 +1,106 @@
...
+        'legalName' => $team->getName(),
+      ]),
+      'startDate' => $start_date,
+    ]);
+
+    $this->stack->queueMockResponse(['team_purchased_plan' => ['purchased_plan' => $purchased_plan->decorated()]]);
+    $purchased_plan->save();
+
+    // Warm the cache for this purchased_plan.
+    $purchased_plan->set('id', $this->getRandomUniqueId());
```
cnovak: This will work for mock client, but what about calling the real Edge API? This seems to set an ID that would not be valid when not using the mock client.

modules/apigee_m10n_teams/tests/src/Traits/ApigeeMonetizationTeamsTestTrait.php
```diff
@@ -0,0 +1,106 @@
...
+   *   The team purchased plan.
+   *
+   * @throws \Drupal\Core\Entity\EntityStorageException
+   * @throws \Twig_Error_Loader
+   * @throws \Twig_Error_Runtime
+   * @throws \Twig_Error_Syntax
+   */
+  public function createTeamPurchasedPlan(Team $team, RatePlanInterface $rate_plan): TeamsPurchasedPlanInterface {
+    $start_date = new \DateTimeImmutable('today', new \DateTimeZone($this->org_default_timezone));
+    $purchased_plan = TeamsPurchasedPlan::create([
```
cnovak: Shouldn't this be using the entity type manager instead of `TeamsPurchasedPlan::create`?

modules/apigee_m10n_teams/tests/src/Traits/ApigeeMonetizationTeamsTestTrait.php
```diff
@@ -0,0 +1,106 @@
...
+      'startDate' => $start_date,
+    ]);
+
+    $this->stack->queueMockResponse(['team_purchased_plan' => ['purchased_plan' => $purchased_plan->decorated()]]);
+    $purchased_plan->save();
+
+    // Warm the cache for this purchased_plan.
+    $purchased_plan->set('id', $this->getRandomUniqueId());
+    $this->stack->queueMockResponse(['team_purchased_plan' => ['purchased_plan' => $purchased_plan->decorated()]]);
+    $purchased_plan = TeamsPurchasedPlan::load($purchased_plan->id());
```
cnovak: Shouldn't this be using the entity type manager instead of `TeamsPurchasedPlan::load()`?

modules/apigee_m10n_teams/tests/src/Kernel/Entity/PurchasedRatePlanEntityKernelTest.php
```diff
@@ -0,0 +1,256 @@
...
+      'team' => $this->team->id(),
+      'product_bundle' => $this->product_bundle->id(),
+      'rate_plan' => $this->rate_plan->id(),
+    ]);
+    static::assertSame("/teams/{$this->team->id()}/monetization/product-bundle/{$this->product_bundle->id()}/plan/{$this->rate_plan->id()}/purchase", $url->toString());
+
+    $request = Request::create($url->toString(), 'GET');
+    $response = $this->container->get('http_kernel')->handle($request);
+
+    $this->setRawContent($response->getContent());
```
cnovak: Is this line needed? I do not see how it is used in rest of method.

modules/apigee_m10n_teams/tests/src/Kernel/Entity/PurchasedRatePlanEntityKernelTest.php
```diff
@@ -0,0 +1,256 @@
...
+    $url = Url::fromRoute('entity.purchased_plan.team_cancel_form', [
+      'team' => $this->team->id(),
+      'purchased_plan' => $this->purchased_plan->id(),
+    ]);
+    static::assertSame("/teams/{$this->team->id()}/monetization/purchased-plan/{$this->purchased_plan->id()}/cancel", $url->toString());
+
+    $request = Request::create($url->toString(), 'GET');
+    $response = $this->container->get('http_kernel')->handle($request);
+
+    $this->setRawContent($response->getContent());
```
cnovak: Same here, is `setRawContent` needed?

cnovak (approved):
Some of these comments may create changes, maybe not, so will approve but expect my questions to be addressed by comments or code changes, thanks

modules/apigee_m10n_teams/tests/src/Traits/ApigeeMonetizationTeamsTestTrait.php
arlina-espinoza: Updated to use the entity type manager.

modules/apigee_m10n_teams/tests/src/Traits/ApigeeMonetizationTeamsTestTrait.php
arlina-espinoza: Updated to use the entity type manager.

modules/apigee_m10n_teams/tests/src/Traits/ApigeeMonetizationTeamsTestTrait.php
arlina-espinoza: Updated to check if the "integration testing" is enabled. If so, the ID isn't set, as the previous save() call would return the entity with an ID.

modules/apigee_m10n_teams/tests/src/Kernel/Entity/PurchasedRatePlanEntityKernelTest.php
arlina-espinoza: Removed :)

modules/apigee_m10n_teams/tests/src/Kernel/Entity/PurchasedRatePlanEntityKernelTest.php
arlina-espinoza: Removed :)

PR #198, opened 2019-09-06
Fix tests because commerce 8.x-2.14 requires profile

arlina-espinoza:
A new release of Commerce ([8.x-2.14](https://www.drupal.org/project/commerce/releases/8.x-2.14)) now requires the profile module, so some tests were failing. This PR fixes that.


Jaesin (approved):
LGTM

PR #216, opened 2020-01-30
[#215] Fix link to add credit to a prepaid balance

### LINKED ISSUE #215: Re-test Add Credit in monetization project. Found an issue when setting up a demo that it doesn't work with the latest commerce module.

arlina-espinoza:
Fixes #215 .
Found an issue where the add credit link was trying to add the credit to the developer by drupal UID instead of by email, so it was failing when querying the API. This PR fixes the link so the job to update the prepaid balance completes successfully.


shadcn (approved):
Looks good. Thanks @arlina-espinoza

PR #222, opened 2020-02-19
[#220] Validate the terms url

### LINKED ISSUE #220: Validate URL in the Terms and Conditions in the Organization Profile
This is a follow up of https://github.com/apigee/apigee-m10n-drupal/issues/205 , where an invalid URL caused http 500 errors upon purchasing plans.

shadcn:
Fixes #220 

This PR tries to create a `Url` from the terms url and if it fails, it fall backs to the supplied one so that the user still sees the checkbox.

Let me know if we should instead show an error for invalid url.


arlina-espinoza (changes requested):
Thanks @arshad . This is a good fallback, but I think we should also log a watchdog error so administrators can know that this needs to be fixed. LMK your thoughts.

shadcn:
@arlina-espinoza Done.

arlina-espinoza (approved):
LGTM. Thanks @arshad :)

PR #232, opened 2020-04-10
[#230] Add Drupal 9 compatibility

### LINKED ISSUE #230: Drupal 9 readiness
Running the module through [mglaman/drupal-check](https://github.com/mglaman/drupal-check) didn't report any deprecation warnings for Drupal 9, so all we need is to mark this module as compatible with D9 is to add `core_version_requirement: ^8 || ^9` to all module info files (https://www.drupal.org/node/3070687).

shadcn:
Fixes #230

### DIFF
--- a/apigee_m10n.info.yml
+++ b/apigee_m10n.info.yml
@@ -4,6 +4,7 @@ package: Apigee
 
 type: module
 core: 8.x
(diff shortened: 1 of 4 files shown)


arlina-espinoza (approved):
Thanks :)

PR #233, opened 2020-04-13
[#227] Implement test payment gateway requirement

### LINKED ISSUE #227: Option to automatically create a test payment gateway
The Add credit submodule has a requirements page that facilitates the configuration of the module. For the payment gateway, it only checks if one is installed or sends you to a page to add a new one. An idea for improvement would be to ask if the user would like to set a default manual test payment gateway, and if so, do it automatically (like the Kickstart installer).

shadcn:
Fixes #227


shadcn:
I have a fix for the failing test in the this PR: https://github.com/apigee/apigee-m10n-drupal/pull/232

arlina-espinoza (changes requested):
This works correctly, however we need to point out that it will "create a manual payment gateway (useful for tests)". 
Perhaps in the form we can add a radio button to allow a user to choose between:
- a) Create a manual payment gateway (useful for tests): and then display the form.
- b) Setup a different payment gateway: if selected, it would redirect them to `/admin/commerce/config/payment-gateways/add`.

The idea here is to make it easier for someone to test out the module (option A), and serve as a checklist for the user that has a more complex config (B).

shadcn:
@arlina-espinoza I've implemented the changes. Let me know what you think.

Requirements_report___Apigee_Kickstart

arlina-espinoza (approved):
LGTM. Thanks @arshad  :)

PR #255, opened 2020-11-05
Run CI tests against Core stable versions.

arlina-espinoza:
Fixes CI tests by using the latest D9 stable release.


shadcn (approved):
Looks good. Thanks for the quick fix @arlina-espinoza

PR #258, opened 2021-02-24
Switching logic between Edge and Apigee X for user menu

phdhiren:
Switching logic between Edge and Apigee X for user menu

### DIFF
--- a/src/EventSubscriber/ValidateMonetizationEnabledSubscriber.php
+++ b/src/EventSubscriber/ValidateMonetizationEnabledSubscriber.php
@@ -67,7 +67,7 @@ class ValidateMonetizationEnabledSubscriber implements EventSubscriberInterface 
   public function validateMonetizationEnabled(GetResponseEvent $event) {
     /** @var \Symfony\Component\Routing\Route $current_route */
     if (($current_route = $event->getRequest()->get('_route_object')) && ($current_route->hasOption('_apigee_monetization_route'))) {
-      if (!$this->monetization->isMonetizationEnabled()) {
+      if (!$this->monetization->isMonetizationEnabled() && !$this->monetization->isOrganizationApigeeX()) {
         $this->messenger->addError(Monetization::MONETIZATION_DISABLED_ERROR_MESSAGE);
       }
     }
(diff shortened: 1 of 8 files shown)


src/EventSubscriber/ValidateMonetizationEnabledSubscriber.php
```diff
@@ -67,7 +67,7 @@ public function __construct(MonetizationInterface $monetization, MessengerInterf
   public function validateMonetizationEnabled(GetResponseEvent $event) {
     /** @var \Symfony\Component\Routing\Route $current_route */
     if (($current_route = $event->getRequest()->get('_route_object')) && ($current_route->hasOption('_apigee_monetization_route'))) {
-      if (!$this->monetization->isMonetizationEnabled()) {
+      if (!$this->monetization->isMonetizationEnabled() && !$this->monetization->isOrganizationApigeeX()) {
```
cnovak, line 70: This logic is saying that all Apigee X orgs are able to be monetized without any other setting, have we confirmed this is true?

src/Monetization.php
```diff
@@ -538,6 +538,14 @@ public function getOrganization(): ?OrganizationInterface {
     return $this->organization;
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function isOrganizationApigeeX(): bool {
+    $org = $this->getOrganization();
+    return ($org && $org->getPropertyValue('features.hybrid.enabled') === 'true');
```
cnovak, line 546: Is this property `features.hybrid.enabled` coming from the actual API response or is this a property name coming from our module or the PHP Client? If this property is coming from PHP Client, we should modify it to be Apigee X since hybrid is a subset of Apigee X.

cnovak: Or is this wrong logic? Should we be looking at the hybrid attribute on the org, or just checking if the Apigee X endpoint is being used? Not all Apigee X orgs are hybrid.

cnovak (commented):
I did not pull down and validate this code, so will let someone else approve.

src/Monetization.php
phdhiren: The property `feature.hybrid.enabled` is coming from actual Apigee API response and not the PHP Client, which is existing differentiation between 4G/5G org response

divya-intelli:
LGTM

divya-intelli (approved):
LGTM

PR #269, opened 2021-04-16
Resolves Kernel test case issues

phdhiren:
Resolves Kernel test case issues for testApiProductAssignmentAccess and testModuleInstall


raakesh-blokhra (approved):
LGTM

PR #274, opened 2021-04-19
Update required version of apigee_client_php and apigee_edge

phdhiren:
Minimum required version of dependent apigee client php library and apigee edge module


raakesh-blokhra (approved):
Looks good

raakesh-blokhra:
Since this is a pre release (as PHP Client as well as APIGEE EDGE) modules are not released, some checks may fail, but once those modules are released, this should work as expected.

PR #285, opened 2021-05-14
Fix unable to re-purchase a rate plan that have just canceled

### LINKED ISSUE #265: Unable to re-purchase a rate plan that I have just canceled
## Description

Unable to re-purchase a rate plan with this module that I have just canceled with end date: now. Although this is possible via API calls. (The Apigee Monetization API backend supports it.)

My gut says the root cause of the issue is [here](https://github.com/apigee/apigee-m10n-drupal/blob/72147d8d41ffe38b513e74847d9c9ffd2de844d6/src/Entity/PurchasedPlan.php#L272-L294) because this method does an incorrect calculation based on dates.

## Steps to Reproduce
1. Go to Prices & Plans and subscribe to a rate plan with start date: today
2. On your Purchased plans page, cancel the subscription with end date: now (or subscribe to an overlapping rate plan which has the same side effect)

## Actual Behavior
1. The cancalled rate plan is visible under "Active and Future Purchased Plans" on "Purchased plans"
2. You cannot subscribe to the rate plan again because the module does not allows it, it says "You have already purchased...".

## Expected Behavior
1. The cancalled rate plan is visible under "Cancelled and Expired Purchased Plans" on "Purchased plans"
2. You can subscribe to the rate plan again, immediatelly - since you have cancelled your subscription with end date: immediatelly.

## Version Info
1.8.0

shishir-intelli:
Closes #265 

Fix "current date-time/today" issue, used for checking "active/ended" users subscription and allows user to immediately cancel the subscription and re-purchase a rate plan that have just canceled.

### DIFF
--- a/src/Entity/PurchasedPlan.php
+++ b/src/Entity/PurchasedPlan.php
@@ -271,14 +271,14 @@ class PurchasedPlan extends FieldableEdgeEntityBase implements PurchasedPlanInte
    */
   public function getStatus(): string {
     $org_timezone = $this->getRatePlan()->getOrganization()->getTimezone();
-    $today = new \DateTime('today', $org_timezone);
+    $today = (new \DateTimeImmutable('now'))->setTimezone($org_timezone);
 
-    // If rate plan ended before today, the status is ended.
+    // If rate plan ended before today or today, the status is ended.
     $plan_end_date = $this->getRatePlan()->getEndDate();
     if (!empty($plan_end_date) && $plan_end_date < $today) {
       return PurchasedPlanInterface::STATUS_ENDED;
     }
-    // If the developer ended the plan before today, the plan has ended.
+    // If the developer ended the plan before today or today, the plan has ended.
     $developer_plan_end_date = $this->getEndDate();
     if (!empty($developer_plan_end_date) && $developer_plan_end_date < $today) {
       return PurchasedPlanInterface::STATUS_ENDED;
--- a/tests/src/Functional/PurchasedPlanStatusTest.php
+++ b/tests/src/Functional/PurchasedPlanStatusTest.php
@@ -1,0 +1,150 @@
+<?php
+
+/**
+ * Copyright 2021 Google Inc.
+ *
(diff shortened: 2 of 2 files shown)


src/Entity/PurchasedPlan.php
```diff
@@ -271,14 +271,14 @@ public function getTermsAndConditions(): bool {
    */
   public function getStatus(): string {
```
mxr576, line 272: Can we have this covered with unit tests also and cover all possible scenarios? Like when the site's timezone is different than Apigee's timezone, etc.

src/Entity/PurchasedPlan.php
```diff
@@ -271,14 +271,14 @@ public function getTermsAndConditions(): bool {
...
     $org_timezone = $this->getRatePlan()->getOrganization()->getTimezone();
-    $today = new \DateTime('today', $org_timezone);
+    $today = (new \DateTimeImmutable('now'))->setTimezone($org_timezone);
```
mxr576, line 274: Ahh, the good old immutable vs mutable object problem.

src/Entity/PurchasedPlan.php
shishir-intelli: Yes definitely. We have road-map for adding various test cases. I will create this as a separate issue so it can be addressed with other test coverage tasks. Thanks @mxr576.

mxr576: I do not see why it should not be added now as there is a known problem with the current implementation, covering the fix with tests before this CR is merged is (IMO) the best way to ensure this bug never occurs again, but ofc, it is up to the team how tasks are being scheduled. :) I am already glad that this issue got a fix.

cnovak: @mxr576 very good point, we will get back to adding/updating tests with a PR to get it accepted. FYI @raakesh-blokhra @giteshk

phdhiren (approved):
LGTM

PR #290, opened 2021-05-25
Removed the patch from composer.json

divya-intelli:
Removed the patch  from composer.json which was added for circleci test to pass.


phdhiren (approved):
LGTM

PR #291, opened 2021-05-26
Fix invalid handling of possible null value

mxr576:
Developer reference can be null a developer-specific rate plan

### DIFF
--- a/src/Entity/Access/RatePlanAccessControlHandler.php
+++ b/src/Entity/Access/RatePlanAccessControlHandler.php
@@ -94,11 +94,14 @@ class RatePlanAccessControlHandler extends EntityAccessControlHandlerBase implem
     // If rate plan is a developer rate plan, and the assigned developer is
     // different from account, deny access.
     if ($rate_plan instanceof DeveloperRatePlanInterface) {
-      if ($developer = $rate_plan->getDeveloper()) {
+      $developer = $rate_plan->getDeveloper();
+      if ($developer) {
         return AccessResult::allowedIf($account->getEmail() === $developer->getEmail())
           ->andIf(AccessResult::allowedIfHasPermission($account, "$operation rate_plan"));
       }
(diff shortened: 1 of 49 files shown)


src/Entity/Access/RatePlanAccessControlHandler.php
```diff
@@ -94,11 +94,14 @@ protected function checkAccess(EntityInterface $entity, $operation, AccountInter
     // If rate plan is a developer rate plan, and the assigned developer is
     // different from account, deny access.
     if ($rate_plan instanceof DeveloperRatePlanInterface) {
-      if ($developer = $rate_plan->getDeveloper()) {
```
mxr576, line 97: Variable assignment in a condition is dangerous and should be avoided.

mxr576: `Error: Call to a member function getEmail() on null in Drupal\apigee_m10n\Entity\Access\RatePlanSubscriptionAccessHandler->access() (line 102 of /mnt/files/local_mount/build/web/modules/contrib/apigee_m10n/src/Entity/Access/RatePlanSubscriptionAccessHandler.php) `

mxr576:
Hi, can somebody review this change?

src/Entity/Access/RatePlanAccessControlHandler.php
mxr576: This means that there was a developer, it had a developer-specific rate plan, but the developer was removed and the developer-specific rate plan remained in the system (Apigee) with a `null` developer reference.

shishir-intelli:
Hi @mxr576,
Could you please resolve the conflicting files.
Same for #292

Thanks

mxr576:
Hi, 
I am not sure why I am going to have time for this on this more than 1-year-old PR, but "Allow edits by maintainers" is checked, so you should be able to rebase my PR in my repo and fix these.

phdhiren:
Closed via #391

PR #298, opened 2021-07-22
ApigeeX code for Cuj7(set default billing of user)

divya-intelli:
This PR updates the billing type of the user as per the default set by the admin also provides a form to the admin to change the billing type if default is not set.


divya-intelli:
Created new PR #312 , hence closing this PR.

PR #300, opened 2021-07-22
ApigeeX code for Cuj6bc(display balance and billing type)

divya-intelli:
This PR displays the prepaid balance and the billing type of the user.


raakesh-blokhra (approved):
LGTM. Tested internally

PR #314, opened 2021-09-16
Default billing type for ApigeeX org.

divya-intelli:
This PR sets the default billing type on user registration for ApigeeX org.


raakesh-blokhra (approved):
LGTM... patches to be removed before we merge main branch

PR #319, opened 2021-10-04
Fix issue apigee_edge_624 - hook function with argument

### LINKED ISSUE #624: Rollback or fix #609 or find an alternative solution
## Description
PR #609 was merged but there are several problems with it.

Approach 1) If we would like to keep the _current_ solution then it should be fixed, see review comments on https://github.com/apigee/apigee-edge-drupal/commit/1f15f9148f15066c6ba7584f8e9c56ac5ce654e7

Approach 2) Since even Drupal core is tending to get away Drupalism called "alter hooks" and use events, the SDKConnector should rather trigger an event instead of exposing an alter hook.

Approach 3) (This would be my preference) instead of extending the public API of the module "unnecessarily", I would verify what changes are needed to be made (if any) to allow a simple SDK connector _service decorator_ to change (extend or override) the user agent string. 

## Version Info
Latest master.

shishir-intelli:
Fix https://github.com/apigee/apigee-edge-drupal/issues/624

Updated the function to use argument for prefixing the user agent string.

### DIFF
--- a/.circleci/RoboFile.php
+++ b/.circleci/RoboFile.php
@@ -492,6 +492,7 @@ class RoboFile extends \Robo\Tasks
 
         // Add rules for testing apigee_edge_actions (only for D8).
         $config->require->{"drupal/rules"} = "3.0.0-alpha6";
+        $config->require->{"drupal/typed_data"} = "1.0.0-beta1 as 1.x-dev";
 
       default:
         break;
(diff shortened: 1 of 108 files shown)


phdhiren (approved):
LGTM

phdhiren (commented):
You need to update minimum required version of apigee_edge

PR #320, opened 2021-10-05
Fix for default billing type for ApigeeX

divya-intelli:
This PR adds the default billing type when the user registers.


raakesh-blokhra (approved):
LGTM

PR #321, opened 2021-10-05
Missed code to add js for ApigeeX

divya-intelli:
This PR adds the js to limit the decimal point in add credit.


raakesh-blokhra (approved):
LGTM. As suggested by Gitesh, we might do some cleanup with method names later if required as a separate PR.

PR #323, opened 2021-10-07
Set current billing type in switch form for Apigee X

divya-intelli:
In this PR by default the current billing type of the developer is selected in billing type switch form.


raakesh-blokhra (approved):
LGTM

PR #324, opened 2021-10-08
Change in violation type for insufficient fund for Apigee X

divya-intelli:
There was a change in violation type in the response due to which the text to add credit to the balance was not shown.This happens when the developer don't have enough balance to buy the api product.


raakesh-blokhra (approved):
LGTM

PR #325, opened 2021-10-14
Changes for billing type switch form and added currency in warning in add credit page.

divya-intelli:
This PR enables the save changes button only when the admin  changes the billing type of the developer also appropriate text will be shown on the confirmation page.This PR also displays the currency along with the wait time on the warning message when the  developer adds credit.


raakesh-blokhra (approved):
LGTM

PR #327, opened 2021-10-21
Updated version of apigee_client_php and apigee_edge

divya-intelli:


composer.json
```diff
@@ -8,10 +8,10 @@
         "ext-bcmath": "*",
         "cweagans/composer-patches": "~1.6",
         "commerceguys/intl": "~1.0",
-        "drupal/apigee_edge": "^1.22",
+        "drupal/apigee_edge": "^2.0.0",
```
phdhiren, line 11: Why not like this "drupal/apigee_edge": "^1.22 || ^2.0.0",

phdhiren: We will have different change with 8.x branch for 2.x above change should be fine.

phdhiren (approved):
LGTM

PR #350, opened 2021-12-07
Add credit functional tests

divya-intelli:


modules/apigee_m10n_add_credit/tests/src/Functional/ApigeeX/AddCreditConfigFormTest.php
```diff
@@ -0,0 +1,126 @@
...
+   * @var \Drupal\commerce_product\Entity\ProductInterface
+   */
+  protected $product;
+
+  /**
+   * {@inheritdoc}
+   *
+   * @throws \Exception
+   */
+  public function setUp() {
```
phdhiren: Can we have void return type as per https://www.drupal.org/node/3114724

phdhiren (approved):
LGTM

PR #366, opened 2022-02-08
Fix for allowing installation of apigee_m10n even if the organization is not monetized

### LINKED ISSUE #365: If organisation is not monetized then should not be able to install apigee-m10n module.
Should not be able to install apigee_m10n module if non monetized organization is used.

divya-intelli:
Fixes #365


phdhiren (approved):
LGTM

PR #395, opened 2022-08-29
$modules is now a protected property in tests

### LINKED ISSUE #394: $modules is now a protected property in tests
## Description
$modules property must be declared protected. See https://www.drupal.org/node/2909426

phdhiren:
Fixes #394


divya-intelli (approved):
LGTM

PR #406, opened 2022-11-15
Fix for dev dependencies error

### LINKED ISSUE #405: Github action fails when installing dev dependencies
## Description
Github action fails when installing dev dependencies as it tries to install both behat/mink-browserkit-driver ^2.1 and friends-of-behat/mink-browserkit-driver ^1.4

divya-intelli:
Fixes #405 
"drupal/drupal-extension" was added when moving from travis to circle ci .
As currently we are using Github action we don't need this library and can be safely removed.
This issue is similar to https://github.com/apigee/apigee-edge-drupal/issues/757

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,6 @@
         "behat/mink-extension": "^2.0",
         "bex/behat-screenshot": "^1.2",
         "drupal/core-dev": "^9.3",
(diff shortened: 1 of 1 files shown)


shishir-intelli (approved):
LGTM

PR #449, opened 2023-09-26
Set default cache value if cache returns NULL while running test case

shishir-intelli:
In Drupal 10.1, fixed test-case failing, due to  error `TypeError: Drupal\apigee_edge_teams\Entity\Team::getCacheMaxAge(): Return value must be of type int, null returned`.
Set default cache value if cache returns NULL while running test case
Originally issue PR https://github.com/apigee/apigee-edge-drupal/pull/917


divya-intelli (approved):
LGTM

PR #513, opened 2025-02-14
Fix added for replacement of Datetime::formatExample()

### LINKED ISSUE #512: Call to deprecated method formatExample() of class Datetime
**Is your feature request related to a problem? Please describe.**
Call to deprecated method formatExample() of class [Drupal\Core\Datetime\Element\Datetime](https://api.drupal.org/api/drupal/core!lib!Drupal!Core!Datetime!Element!Datetime.php/class/Datetime/10.4.2.x). Deprecated in drupal:10.2.0 and is removed from drupal:11.0.0. There is no replacement.

**Describe the solution you'd like**
Remove the implementation of formatExample() and change it with static text YYYY-MM-DD hh:mm:ss

kedarkhaire:
Fixes #512 

@deprecated in drupal:10.2.0 and is removed from drupal:11.0.0. 
HTML date input format should not be exposed to users, as browser widgets expose input differently, varying by vendor, locale, device type, etc.
https://www.drupal.org/project/drupal/issues/2791693

### DIFF
--- a/src/Plugin/Field/FieldWidget/DatestampWidget.php
+++ b/src/Plugin/Field/FieldWidget/DatestampWidget.php
@@ -38,11 +38,12 @@ use Drupal\Core\Form\FormStateInterface;
 class DatestampWidget extends TimestampDatetimeWidget {
 
   /**
-   * Example format used as Datetime::formatExample is removed from D11.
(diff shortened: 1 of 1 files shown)


divya-intelli (approved):
LGTM

PR #527, opened 2025-03-24
Replace xdebug with pcov for code coverage tool

### LINKED ISSUE #525: Segmentation fault when using --coverage-clover
## Description
The fact that the segmentation fault occurs specifically when using --coverage-clover strongly points to Xdebug as the primary suspect.
Xdebug, especially with code coverage enabled, adds significant memory overhead. It instruments the code execution to track which lines are executed, which uses a lot of memory.
Code coverage dramatically slows down test execution.
Xdebug versions can sometimes have compatibility issues with specific PHP and PHPUnit versions, leading to crashes.

## Solution
As Xdebug cause the issues, we can use alternative code coverage tool - **pcov** which is a lightweight code coverage driver that is significantly faster and uses less memory than Xdebug.

## Resource
https://xdebug.org/reporting-bugs
https://thephp.cc/articles/pcov-or-xdebug?ref=phpunit

kedarkhaire:
Fixes #525

### DIFF
--- a/.github/workflows/php.yml
+++ b/.github/workflows/php.yml
@@ -67,7 +67,7 @@ jobs:
     - name: "Install PHP"
       uses: "shivammathur/setup-php@v2"
       with:
-        coverage: "xdebug"
+        coverage: "pcov"
         php-version: "${{ matrix.php-version }}"
         tools: composer:v2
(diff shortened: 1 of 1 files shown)


divya-intelli (approved):
LGTM

PR #538, opened 2025-04-17
Fixes hook_requirement install issue on only M10n install

kedarkhaire:
Fixes WSOD on only M10n install when Apigee module is not installed & the GCP credentials are not configured.
Referring document - https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Extension%21module.api.php/function/hook_requirements/11.x

### DIFF
--- a/apigee_m10n.install
+++ b/apigee_m10n.install
@@ -89,6 +89,8 @@ function apigee_m10n_requirements($phase) {
         // If not able to connect to Apigee Edge.
         $monetized = FALSE;
       }
+    } else {
+      $monetized = FALSE;
     }
   }
 
--- a/apigee_m10n.install
+++ b/apigee_m10n.install
@@ -34,70 +34,77 @@ use Drupal\user\Entity\Role;
  */
 function apigee_m10n_requirements($phase) {
   $requirements = [];
+  if (Drupal::moduleHandler()->moduleExists('apigee_edge')) {
+    // If monetization is disabled after module has been installed, show error on status page.
+    if ($phase === 'runtime') {
+      try {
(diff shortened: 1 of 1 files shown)


apigee_m10n.install
```diff
@@ -68,36 +67,37 @@ function apigee_m10n_requirements($phase) {
...
-      /** @var \Drupal\apigee_edge\SDKConnectorInterface $sdk_connector */
-      $sdk_connector = \Drupal::service('apigee_edge.sdk_connector');
-      $org_controller = \Drupal::service('apigee_edge.controller.organization');
-      /* @var \Apigee\Edge\Api\Management\Entity\Organization $organization */
-      $organization = $org_controller->load($sdk_connector->getOrganization());
-
-      // Check if org is ApigeeX.
-      if ($organization && ('CLOUD' === $organization->getRuntimeType() || 'HYBRID' === $organization->getRuntimeType())) {
-        if (!$organization->getAddonsConfig() || !$organization->getAddonsConfig()->getMonetizationConfig() || (FALSE === $organization->getAddonsConfig()->getMonetizationConfig()->getEnabled())) {
+    if (Drupal::moduleHandler()->moduleExists('apigee_edge')) {
```
shishir-intelli, line 70: This module depends on the Edge module, so Edge will be installed automatically. Given that, why is this condition necessary?
Also, this condition implies that the monetization error won’t be shown if the M10N module is installed and the Monetization organization is used.

kedarkhaire: Hi Shishir, Updated the condition, for same error message.

divya-intelli (approved):
LGTM

PR #539, opened 2025-04-23
Fixes M10n independent installation

### LINKED ISSUE #541: Apigee Monetization throws error in independant installation
## Description
In Drupal 11, When we try to install Apigee Monetization & if any of its dependencies are not enabled, then we get WSOD.

## Actual Behavior
Throws WSOD

Should open confirm Module list page asking other dependent modules will be installed too - as a Drupal behaviour.

## Version Info
Drupal 11

kedarkhaire:
Fixes #541

### DIFF
--- a/apigee_m10n.install
+++ b/apigee_m10n.install
@@ -34,7 +34,6 @@ use Drupal\user\Entity\Role;
  */
 function apigee_m10n_requirements($phase) {
   $requirements = [];
-
   // If monetization is disabled after module has been installed, show error on status page.
   if ($phase === 'runtime') {
     try {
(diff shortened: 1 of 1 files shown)


apigee_m10n.install
```diff
@@ -67,39 +67,35 @@ function apigee_m10n_requirements($phase) {
   if ($phase === 'install') {
     $message = t("Apigee Monetization module functionality is not available for the organization used.");
     $monetized = TRUE;
-    if (Drupal::moduleHandler()->moduleExists('apigee_edge')) {
```
giteshk, line 70: Could we keep this check     if (Drupal::moduleHandler()->moduleExists('apigee_edge')) {

kedarkhaire: Okay I will update and check if it works the same as we want.

divya-intelli (approved):
LGTM

PR #604, opened 2026-01-20
Fix phpunit utf8 case-sensitive mismatch

### LINKED ISSUE #605: Fix UTF-8 case-sensitivity mismatch in Header assertions
### Description
This issue addresses a failure in the PHPUnit test suite regarding HTTP header validation. Currently, there is a mismatch when handling UTF-8 characters in headers due to case-sensitivity issues, causing false negatives in the test results.

### Problem
The current test assertion expects a specific casing for UTF-8 header values (or keys), but the actual output differs in case while remaining semantically identical. This leads to brittle tests that fail despite the logic being correct.

shailesh-google:
Fix: #605


divya-intelli (approved):
LGTM

PR #609, opened 2026-02-10
Modify ChromeDriver args to fix D10 Curl/Memory error

### LINKED ISSUE #608: Chrome Crash (CurlExec) in D10 vs Connection Refused in D11
We are experiencing a "matrix hell" scenario in our GitHub Actions workflow. The configuration required to make Drupal 11.2 pass causes Drupal 10.5 to fail with browser crashes, and the configuration that stabilizes Drupal 10.5 causes Drupal 11.2 to fail with connection errors.

1. The Error in Drupal 10.5
When running tests on Drupal 10.5, we encounter a WebDriver\Exception\CurlExec. This indicates that the ChromeDriver/Chrome process is crashing mid-test (likely due to /dev/shm exhaustion).
Drupal\Tests\apigee_m10n_add_credit\FunctionalJavascript\ApigeeX\AddCreditCustomAmountTest::testUnitPriceValidation
WebDriver\Exception\CurlExec: Curl error thrown for http GET to http://127.0.0.1:9515/...
...
{"message":"stale element reference: stale element not found...

2. The Error in Drupal 11.2
When running the same workflow on Drupal 11.2, the tests fail immediately because Mink cannot connect to ChromeDriver. This is due to stricter port/IPv6 binding in newer Drupal/PHPUnit dependencies requiring an explicit port definition.
1) Drupal\Tests\apigee_m10n_add_credit\FunctionalJavascript\AddCreditCustomAmountTest::testPriceRangeField
Behat\Mink\Exception\DriverException: Could not open connection: Failed to connect to 127.0.0.1 port 9515 after 0 ms: Couldn't connect to server

**Root Cause Analysis**

**D10 Failure**: The default Docker/CI shared memory (/dev/shm) is 64MB. Complex JS tests in D10 exhaust this, causing Chrome to crash.

**D11 Failure**: Newer Mink/ChromeDriver versions do not bind to 127.0.0.1:9515 by default (often defaulting to IPv6 or a socket), causing the "Connection Refused" error when Drupal tries to connect.

shailesh-google:
Fix #608

### DIFF
--- a/.github/workflows/php.yml
+++ b/.github/workflows/php.yml
@@ -14,7 +14,7 @@ env:
   APIGEE_EDGE_HYBRID_ORGANIZATION: ${{ secrets.APIGEE_EDGE_HYBRID_ORGANIZATION }}
   BROWSERTEST_OUTPUT_DIRECTORY: "sites/simpletest/browser_output"
   BROWSERTEST_OUTPUT_BASE_URL: ""
-  MINK_DRIVER_ARGS_WEBDRIVER: '["chrome", { "chromeOptions": { "w3c": false } }, "http://127.0.0.1:9515/wd/hub"]'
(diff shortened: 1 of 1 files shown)


divya-intelli (approved):
LGTM

PR #624, opened 2026-06-02
update PriceRange element to extend FormElementBase

shishir-intelli:
Updated the PriceRange custom form element plugin to inherit from FormElementBase instead of FormElement. This resolves class compatibility errors in newer versions of Drupal where FormElement behavior has changed or 
strict typing expects FormElementBase for standard form inputs.

- Updated namespace use statement to FormElementBase
- Updated PriceRange class definition to extend FormElementBase


divya-intelli (approved):
LGTM

